PR #49, opened 2026-04-12
fix: add WINBOOL type

gynt:
Adds missing WINBOOL type in a predictable location

### DIFF
--- a/src/precomp/common.hpp
+++ b/src/precomp/common.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+typedef int WINBOOL;
+
 namespace Ghidra {
 typedef void* pointer;
 typedef void* pointer32;


src/precomp/common.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+typedef int WINBOOL;
```
TheRedDaemon, line 3: OK, I am curious:  
This one is not part of the Windows.h header?
From where is it and where do we need it?

TheRedDaemon (changes requested):

src/precomp/common.hpp
gynt: Exported Direct x 8 stuff needs it (function signatures). It should be in Windows psapi.h but I couldn't find it. Therefore this stub.
I remember now, WINBOOL is in psapi.h in windows 10 SDK but not in the msvc1400 SDK. WINBOOL doesn't exist in MSVC1400 for some reason while it should. I also can't trace it using the dplay.h headers I found online.
Funnily enough, our dplay.h from 2007 just uses BOOL... I guess this is a non-issue and should be fixed in Ghidra instead.

PR #50, opened 2026-04-12
remove ADDRESS_ prefix from address enum

gynt:


src/precomp/addresses-SHC-3BB0A8C1.hpp
TheRedDaemon, line 1: You need to adjust the script in [extract_header_and_update.py](https://github.com/sourcehold/OpenSHC/blob/refactor/address-enum-style-without-address-prefix/status/extract_header_and_update.py) and replace the current export txt with a new one.

I used an regex to get the addresses from the header. 😄

TheRedDaemon (changes requested):

src/precomp/addresses-SHC-3BB0A8C1.hpp
gynt: Done!

TheRedDaemon (approved):

PR #51, opened 2026-04-30
release of initial hpp files

gynt:


gynt:
I forgot the clang-format... Sorry about that. Will fix. But in theory, does this look good to you?

TheRedDaemon:
It looks good. I would just like to understand the general structure:
- Object-Instances (STRUCT_RESOLVER) are all placed in "Globals"?
- The FUNCTION_RESOLVERS are all placed next to their class headers?

My head is debating whether or not the FUNCTION_RESOLVERS are good to be "Mixed" with the headers. Although, I guess this will one day make it easier to switch the "normal" class headers.

One questions:
- Where will the implementation files go?

gynt:
> Object-Instances (STRUCT_RESOLVER) are all placed in "Globals"?

Yes, although Ghidra allows moving global labels into namespaces, I am not very keen on doing that because it makes decompiled code more verbose. So defined data with a symbol (struct resolver) will go into Globals. I tried to identify the major structures as much as I could, but there still could be some gains to be made (nesting structs) in the future.

> The FUNCTION_RESOLVERS are all placed next to their class headers?

Yes all placed next to the key .hpp files. This makes it easy to find them (to change the "implemented" parameter to true). They are all marked with `.func.` so easy to find and delete/move, etc.

> Where will the implementation files go?

Method reimplementations go into a subfolder with the same name as the class:
```
Class.hpp
Class.func.hpp
Class/methodA.cpp
Class/methodB.cpp
Class/methodC.cpp
```

Function reimplementations go into the same folder as the namespace.hpp file: (but I guess this can be made identical to the method reimplementation)
```
Namespace/
Namespace/Namespace.hpp
Namespace/Namespace.func.hpp
Namespace/funcA.cpp
Namespace/funcB.cpp
Namespace/funcC.cpp
```

I could add an example already but then I will change the current one for xyAreValid etc. Is that okay with you?

TheRedDaemon:
Ah, so they will be mixed inside the folder structure. I guess that makes sense.

I would say, you rather finish this PR (fix formatting) and add another one with one, two implementation examples.

gynt:
> Ah, so they will be mixed inside the folder structure. I guess that makes sense.
> 
> I would say, you rather finish this PR (fix formatting) and add another one with one, two implementation examples.

Yes with the key benefit being that every file has a predictable location.

TheRedDaemon (approved):
All discussed. Failing linter accepted for this PR, since too big.
Should we run more often into size limitations, we might need to re-think the linter task.

PR #53, opened 2026-05-01
resolve #52: include strings header file

### LINKED ISSUE #52: Handle literal strings better
The original binary has strings that were used as literal strings stored in a particular section of the binary. To achieve 100% via `reccmp-reccmp`, we need to invoke `STRUCT_RESOLVER` like so:
```cpp
MACRO_STRUCT_RESOLVER(char const[], false, (Address::SHC_3BB0A8C1_0x005A6CC4)) pWAV;

assert *pWAV::ptr == ".wav";
```
This is a shortcoming of `reccmp-reccmp`, it could easily check if the string from the source code is the same as the string in the original binary, and then consider the match 100%.

See example:
Image

If this is addresssed, the address enum file can be reduced as it doesn't need addresses of literal strings anymore.

## Further thoughts:
This doesn't work either, the reccmp isn't happy.
```cpp
char const* const WAV = ".wav";
```
So currently, reccmp is only happy about the STRUCT_RESOLVER method.
It does get 100% if this is used: 
```cpp
// STRING: STRONGHOLDCRUSADER 0x005A6CC4
char const* const WAV = ".wav";
```

Which leaves us with the conundrum of naming the strings.

gynt:
Resolves #52 by exporting literal strings and their address from Ghidra into a file, listing strings by address. Ready for usage in OpenSHC by including said file.

### DIFF
--- a/tools/mcp/ghidra_scripts/stringexporter.py
+++ b/tools/mcp/ghidra_scripts/stringexporter.py
@@ -1,5 +1,6 @@
-#TODO write a description for this script
-#@author 
+#This script exports .rdata literal strings and addresses into a file. 
+# Labels are either default Ghidra labels or labels set by users.
+#@author Gynt
 #@category _OPENSHC.TOOLS.DECOMPILATION
 #@keybinding 
 #@menupath 
@@ -7,7 +8,6 @@
 #@runtime Jython
 
 
-#TODO Add User Code Here
 
 import re
 PATTERN = re.compile("[^a-zA-Z0-9_]+")
--- a/tools/mcp/ghidra_scripts/stringexporter.py
+++ b/tools/mcp/ghidra_scripts/stringexporter.py
@@ -1,0 +1,31 @@
+#TODO write a description for this script
+#@author 
+#@category _OPENSHC.TOOLS.DECOMPILATION
+#@keybinding 
+#@menupath 
+#@toolbar 
+#@runtime Jython
+
+
+#TODO Add User Code Here
+
+import re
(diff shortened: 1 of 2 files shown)


tools/mcp/ghidra_scripts/stringexporter.py
```diff
@@ -0,0 +1,31 @@
+#TODO write a description for this script
```
TheRedDaemon: Still in this PR?

tools/mcp/ghidra_scripts/stringexporter.py
```diff
@@ -0,0 +1,31 @@
...
+#@author 
+#@category _OPENSHC.TOOLS.DECOMPILATION
+#@keybinding 
+#@menupath 
+#@toolbar 
+#@runtime Jython
+
+
+#TODO Add User Code Here
```
TheRedDaemon: Still in this PR? Or what is this header part about?

TheRedDaemon (changes requested):
Also, what about the TODO in the PR description?

tools/mcp/ghidra_scripts/stringexporter.py
gynt: I will delete, it is created by Ghidra.

tools/mcp/ghidra_scripts/stringexporter.py
gynt: Yes, good idea. Fixed it!

TheRedDaemon (approved):

PR #54, opened 2026-05-02
[CORE] Remove test files

TheRedDaemon:
This also marks the end for a compilable EXE target for a loooong time, since it will only ever build if all resolvers included via headers are implemented or properly instantiated.

Also, `core` is now down to one file... well, but it needs a place. 🙃


gynt:
bye bye EXE build :)

src/entry.cpp
```diff
@@ -13,7 +13,7 @@ extern "C" __declspec(dllexport) int __cdecl luaopen_OpenSHC(lua_State* L) { ret
 
 int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR lpCmdLine, int nCmdShow)
 {
-    return MACRO_CALL(Main_Func::WinMain)(hInstance, hPrevInstance, lpCmdLine, nCmdShow);
+    return MACRO_CALL(OpenSHC::Global_Func::WinMain)(hInstance, hPrevInstance, lpCmdLine, nCmdShow);
```
gynt, line 16: The windows spec technically doesn't allow calling a WinMain ourselves. Just a thing to keep in mind if this doesn't turn out to work!

gynt (approved):

PR #56, opened 2026-05-03
[EXPERIMENT/MSS] Attempt func 0x00467810 and 0x004799A0

TheRedDaemon:
Run into multiple things that might need updates on Ghidra side. Here my comments from one file:
> // TODO:
// - DAT_00df3844 did not exist and I created it to test it
// - the mss types need to use the mss32 header, not the Ghidra exported one (how to handle this?)
// - "_close" is treaded like an OFFSET function and reduces the score of reccmp
// - "UnkSoundFlagsAndLoopCount" plays not nice with any number-like operation, the more likely thing therefore are
// either a union, which allows to access the different parts separately, but this would likely compile to WORD
// operations, or a bitfield, maybe like this:
//      int loopCount : 16;
//      int reserved : 13;
//      int unknownFlag1 : 1;
//      int unknownFlag2 : 1;
//      int uninterruptable : 1; // ?
// In a test in the current file it compiled to "& 0x80000000" properly when using the specific field, but no idea if
// this works for multifield checks or assigns

EDIT:
Currently have an assertion error if the "_close" is called. Something does not fit at the moment...


gynt:
> - the mss types need to use the mss32 header, not the Ghidra exported one (how to handle this?)

I tweak this in the Ghidra export such that all files that use Mss32 use Mss32.h. Basically I collapse everything underneath a certain namespace/folder in Ghidra to the Mss32.h header. Just like it is done for Directplay.

## Regarding _close
> "_close" is treaded like an OFFSET function and reduces the score of reccmp

Can you show the dump / diff?
Have you tried? https://github.com/isledecomp/reccmp/blob/master/docs/annotations.md#annotating-a-comment-of-the-functions-symbol
Or: https://github.com/isledecomp/reccmp/blob/master/docs/annotations.md#library

> Currently have an assertion error if the "_close" is called. Something does not fit at the moment...

We need to make sure we use MSVCRT not UCRT (I had issues with mixing the two). Or perhaps it is the other way around and we have to use UCRT.
Also, we may need this compiler flag: https://github.com/UnofficialCrusaderPatch/UnofficialCrusaderPatch3/issues/58

Would it be better if we implement _close via a function resolver macro call? Then we use the game's own one for the time being, avoiding any IO issues. Shall I use `OpenSHC::IO::WindowsHelpers` as a namespace for _close, _open, etc.?

TheRedDaemon:
We might end up needing these CRT functions as resolvers, yes. Likely with a shared "implemented" flag. With a bit of luck we maybe can just use the ptr to the windows version for "implemented".
The important thing seems to be this:
[Link](https://learn.microsoft.com/hi-in/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-160)

If Crusader has the CRT statically linked, we are out of luck, and need to hook the functions anyway. It might actually be, because this might have been the reason I switched our build to static CRT a while ago, but I have not checked the dll imports in a while.

**Regarding the Mss32 header**
Bink would be another case where I created "Fake" Structures. I mean, I did again in our bink-header, but they are different. 🙃

gynt:
Please try #57 and see if it addresses your mss32 issues!

src/OpenSHC/OS/OS.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include "OpenSHC/OS.hpp"
+
+#include "float.h"
+#include "io.h"
+
+// NOTE: Activating them in DLL mode currently results in heap corruption
+
+// this one is hard to pin down, maybe an STL this-function, or a special buffer?
+// STUB: STRONGHOLDCRUSADER 0x004791B0
+// int* __stdcall basic_ofstream_write(void* param_1, uint param_2) {TODO}
```
gynt: Probably
```cpp
std::basic_ofstream<char, std::char_traits<char>>::write
```
As in:
```cpp
std::ofstream file("screenshot.bmp", std::ios::binary);
file.write(buffer, size);
```

TheRedDaemon: But no this pointer? Or is there still something missing?
Or maybe a `<<`?

The streambuffer is actually a real annoyance, since it might be the only "complicated" structure with vtables and such. Hm, real hard to judge.

gynt: I think we are allowed to stub it :) We know what it is, what it is doing, but we won't be able to clearly reimplement it I think.

yes it could be a `<<`, you are right. Then first "this" argument is the filestream. And I guess C++ features detect how long the buffer is? Otherwise there needs to be another argument (length of buffer). Oh wait, perhaps a bmp struct is passed in with a known length.

TheRedDaemon:
Hi @gynt , I still have some mismatches with main:
- Was 0x00df3844 really added somewhere in the Ghidra export?
- `__nan` was removed, but this is something directly callable.
- `__alloca_probe` is still there, but is this somthing callable? I do not find any MS docs at least.
- There is a mismatch compared to the last status export on main.
- "_toupper" and "_tolower" might need a renaming in our code: Windows.h has Macros defined for them.
- I am not sure if we want to hook three free functions, cosidering two of them jump to the third. Regardless having three with the same name will not work anyway.

gynt:
> Was 0x00df3844 really added somewhere in the Ghidra export?

Yes, that address is part of Structure `SoundEffectsHelperData1`. Globally, this is `DAT_SoundEffectsHelperData1::ptr->field14_0x54`.
https://github.com/sourcehold/OpenSHC/blob/85aef21e7668d678006a0c4c7afdacbe067eac50/src/OpenSHC/Audio/SoundEffectsHelperData1.hpp#L40

> __nan was removed, but this is something directly callable.

You mean `__isnan`? That function is only called by `math_atan_FUN` functions, so it wasn't called by non-lib code.

> __alloca_probe is still there, but is this somthing callable? I do not find any MS docs at least.

It is also known as `__chkstk` (https://learn.microsoft.com/en-us/windows/win32/devnotes/-win32-__chkstk). The compiler should invoke it automatically. Ghidra doesn't export it literally but always inlines it into the decompilations. I am not sure how to handle this one at this point, because the compiler includes it by itself when needed, so it behaves like the injected math lib functions I guess?

> "_toupper" and "_tolower" might need a renaming in our code: Windows.h has Macros defined for them.

I renamed them to `__toupper` and `__tolower` upstream in Ghidra (and also renamed the three `free` variations with their true names: `_free_base` and `_free` are called by non-lib functions so they need to be exported). I haven't updated this yet in main. I can do this so you can use it.

src/OpenSHC/OS/OS.cpp
TheRedDaemon: I think we postpone the discussion regarding the bytestream to when it is needed.

TheRedDaemon:
The only remaing things are the following three:
- I would say we trust the compiler when is comes to `__alloca_probe `. I think it should be remove for now.
- `sprintf` needs the varargs declaration
- Is it possible for `localtime` to use the Windows neutral variant, like in my changes? I do not entirely trust the choice of Ghidra here and would like the compiler and the environment to decide this.

TheRedDaemon:
@gynt I still had same requests here. No idea if all of them are as easy.

gynt:
All points should have been addressed in 58806bb0579cb9a0547ba0e32b6cbfa16994c93e (took a while because the sarif format changed in ghidra 12)

gynt (approved):

PR #57, opened 2026-05-06
refactor: exclude bink and mss32 fake files, use include of header

gynt:
Addresses the miles (and bink issues) mentioned in #56 

Note namespace `mss` changed to `MSS`, `SND_SoundSystemObj` changed to `DAT_SoundSystemState`
Namespace `_enums` changed to `enums`


TheRedDaemon:
Ok, I am a bit confused. Are the fake definition not in scope due to this changes, or do I need to test this?

Could it be possible to just import our Fake headers from the project here into Ghidra without adding it to the export, use the structs and values from there and remove the Fake structure inside the Crusader folder in Ghidra completely?

gynt:
Hmmm, what I wanted to happen is that those Fake files are deleted, as including Mss32.h would be enough to cover all relevant reimplementations, right?

So everything in the /Mss32/ folder in this PR can be removed, right?

So, for example, this usage is now correct, right?
https://github.com/sourcehold/OpenSHC/pull/57/changes#diff-e8268c733d50a21e95087c973d8cd5fb1770c98832895765310f563ca0594571R10-R14

gynt:
I cleaned up the bink and miles files (and directplay files) we don't need from the export from Ghidra as we have sufficient header files ourselves. Can you confirm this PR works and solves your issue? @TheRedDaemon 

See previous comment clarifying what is happening.

TheRedDaemon (approved):
Looks good. I will merge it and try it in the other branch.

PR #58, opened 2026-05-09
Refactor windows lib functions

gynt:
Moved os-specific functions that the game uses to OpenSHC/OS instead of leaving them out of the repo. This allows us to get 100% match and avoid crashes due to mixing statically linked CRT in the original exe and in the OpenSHC dll.

`REIMPLEMENTED_CRT` is set to 0 in CMakeLists.txt for the OpenSHC dll target. This should be set to 1 if all OS functions are reimplemented (or stubbed) and testing is required.

We can either use `// STUB` or `// FUNCTION` for any OS function depending on whether we want to ignore reimplementation accuracy of the static lib function (preferred), or whether we want to actually track reimplementation progress of the lib function (can be used if static lib functions turn out to be linker/compiler generated in a custom fashion for the original binary)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,7 +163,7 @@ target_file_copy_if_different(OpenSHC.exe.runnable "${CRUSADER_DIR}/shfolder.dll
 add_library(OpenSHC.dll SHARED src/entry.cpp ${CORE_SOURCES} ${OPENSHC_SOURCES})
 set_target_properties(OpenSHC.dll PROPERTIES OUTPUT_NAME ${OPEN_SHC_NAME})
 set_target_properties(OpenSHC.dll PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/DLL")
-target_compile_definitions(OpenSHC.dll PRIVATE OPEN_SHC_DLL)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -163,7 +163,7 @@ target_file_copy_if_different(OpenSHC.exe.runnable "${CRUSADER_DIR}/shfolder.dll
 add_library(OpenSHC.dll SHARED src/entry.cpp ${CORE_SOURCES} ${OPENSHC_SOURCES})
 set_target_properties(OpenSHC.dll PROPERTIES OUTPUT_NAME ${OPEN_SHC_NAME})
 set_target_properties(OpenSHC.dll PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/DLL")
-target_compile_definitions(OpenSHC.dll PRIVATE OPEN_SHC_DLL)
+target_compile_definitions(OpenSHC.dll PRIVATE OPEN_SHC_DLL REIMPLEMENTED_CRT=0)
```
TheRedDaemon, line 166: Did you test setting this 1 and import the header somewhere, just to test if it works?

TheRedDaemon (changes requested):

CMakeLists.txt
TheRedDaemon: Thought about it, and I will test it anyway in the other branch, so this is fine. 👍

TheRedDaemon (approved):

PR #62, opened 2026-05-12
reimplement: SHC_3BB0A8C1_0x004CB060 100%

gynt:
This reimplementation of `OpenSHC::AI::AICState::setFoodBuyPlan` is 100% matching with the original binary.


src/OpenSHC/AI/AICState.func.hpp
```diff
@@ -602,8 +602,8 @@ namespace AI {
             void (AICState::*)(int), false, Address::SHC_3BB0A8C1_0x004D3780, &AICState::computeNervousness)
         computeNervousness;
 
-        MACRO_FUNCTION_RESOLVER(undefined4 (AICState::*)(int, int), false, Address::SHC_3BB0A8C1_0x004D3850,
-            &AICState::determineAIPlayerHelp)
+        MACRO_FUNCTION_RESOLVER(
+            BOOLEnum (AICState::*)(int, int), false, Address::SHC_3BB0A8C1_0x004D3850, &AICState::determineAIPlayerHelp)
```
gynt, line 606: This is actually an upstream update from the database. I can exclude if it you want, but it won't really matter or cause conflicts I think with main.

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
```diff
@@ -0,0 +1,71 @@
...
+namespace AI {
+
+    using OpenSHC::Game::Resources::ResourceType;
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x004CB060
+    void AICState::setFoodBuyPlan(int playerID)
+    {
+        int aiType;
+        int aiConfigIndex;
+        int minimumStock;
```
TheRedDaemon: I think for good form, moving local variable definitions to their first usage and setting `const` when applicable would be a good thing. The early definition in Ghidra at least would not be like I write code.

Unless of course it messes with the byte results.

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        aiType = DAT_GameState::ptr->playerDataArray[playerID].aiType;
+
+        // Early return if not an AI player (human player has aiType == 0)
+        if (aiType == 0) {
```
TheRedDaemon: We have not constants/enums for the ai type, right?

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
```diff
@@ -0,0 +1,71 @@
...
+            return;
+        }
+
+        // Get AI configuration index (convert 1-based to 0-based index)
+        aiConfigIndex = aiType - 1;
+
+        // Check each food type and queue purchase if current stock is below minimum
+
+        // Apples: special handling with >= 0 check (allows -1 to disable)
+        minimumStock = this->DAT_AICArray[aiConfigIndex].minimumApples;
```
TheRedDaemon: Can you check if splitting the locals works? Currently the compiler re-uses the same value. I wonder if a `const minimumApples` would be folded together with a `const minimumChesse` easily.

That is just what I quickly could think of. Out of pure habit I would likely store the long "currentResources" value in locals also (or refactor to a function). But I can imagine that this might be the point where the compiler might rebel, since the access needs to happen after the minimum stock check...

Also, why the special apple handling?

src/OpenSHC/AI/AICState.func.hpp
TheRedDaemon: Should be fine. You create the exports after all.

TheRedDaemon (changes requested):

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
gynt: Addressed

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
gynt: Addressed (we do have them)

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
gynt: This works: `int* currentResources = DAT_GameState::ptr->playerDataArray[playerID].currentResources;`
And also setting a variable for every resource type works. `const`  isn't necessary because the compiler is smart enough to realize this.

So, addressed!

TheRedDaemon: A very nice with the temporary.

I assume these resource indexes are also used in other positions? Else I would be wondering if the `currentResources` might just be a big struct.

Also: `const` in this cases is not for the compiler, default-const has become basically a standard in many more modern languages. So it is just another code-style remark. 😅

TheRedDaemon (changes requested):

src/OpenSHC/AI/AICState/setFoodBuyPlan.cpp
TheRedDaemon: Was done. 👍

TheRedDaemon (approved):

PR #63, opened 2026-05-12
reimplement: SHC_3BB0A8C1_0x004ECEF0 100%

gynt:
This function `AIVState::setupAIVMetadata` is reimplemented 100% matching with original binary (which required reimplementing `AICState::getAIBuildInterval` which will be pushed in a different PR).


src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace OpenSHC {
+namespace AI {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x004ECEF0
+    int AIVState::setupAIVMetadata(int playerID)
+    {
+        int aivID = 1;
+        AIVSpec* pSlot = &this->SEC_AIVS[1];
+        do {
```
TheRedDaemon: Have you checked if this can also be a classic "for" loop?
I currently do not know if Ghidra generates them, but 1 to < 9, with the two values getting ++, feels like it.
Could it also be that `aivID` is the iteration number and AIVSpec ref is accessed once inside the loop, maybe even as reference and not ptr (`AIVSpec& pSlot = this->SEC_AIVS[aivID]`)?

For me it just does not make much sense to write such code. Maybe this could also be a hint for the AI, so see if there are loop variables that go together.

src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
```diff
@@ -0,0 +1,39 @@
...
+            if (pSlot->playerID == 0) {
+                this->SEC_AIVS[aivID].playerID = playerID;
+
+                int aiType = DAT_GameState::ptr->playerDataArray[playerID].aiType;
+                this->SEC_AIVS[aivID].aiType = aiType;
```
TheRedDaemon: Direct assign possible?

src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace OpenSHC {
+namespace AI {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x004ECEF0
+    int AIVState::setupAIVMetadata(int playerID)
+    {
+        int aivID = 1;
+        AIVSpec* pSlot = &this->SEC_AIVS[1];
+        do {
+            if (pSlot->playerID == 0) {
```
TheRedDaemon: Does a if-not-equal-continue produce the same code?
It might even be possible to use `if (pSlot->playerID)`, but this might be not telling.
Is the player id any kind of enum, or are these just the "slots"? If these really are just numbers and the "0" is like "nothing", a `if (pSlot->playerID)` might be fitting.

src/OpenSHC/AI/AIVState.hpp
```diff
@@ -10,25 +10,25 @@
...
     using OpenSHC::AI::AIGridComputationResourceTypeInt;
     using OpenSHC::AI::AIMapExtraInfo;
-    using OpenSHC::AI::AIV::AIV;
     using OpenSHC::AI::AIV::AIVBuildingType2Int;
     using OpenSHC::AI::AIV::AIVBuildingType2Short;
+    using OpenSHC::AI::AIV::AIVSpec;
     using OpenSHC::Commands::CommandBuildingTypeInt;
     using OpenSHC::Game::Player::PlayerID;
-    using OpenSHC::Map::HeatMap1114;
+    using OpenSHC::Map::HeatMap;
```
TheRedDaemon, line 31: Are these things you already integrated in Ghidra?
EDIT: Switching to this branch actually does not compile this file, since the HeatMap header is missing?

TheRedDaemon (changes requested):

src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
TheRedDaemon: Out of curiosity, I attempted to add the things I mentioned. This one produced 97.44% (only the macro differs, since not implemented):
```cpp
for (int aivID = 1; aivID < 9; ++aivID) {
    if (this->SEC_AIVS[aivID].playerID != 0) {
        continue;
    }

    this->SEC_AIVS[aivID].playerID = playerID;

    this->SEC_AIVS[aivID].aiType = DAT_GameState::ptr->playerDataArray[playerID].aiType;

    this->SEC_AIVS[aivID].currentStepGoal = 0;
    this->SEC_AIVS[aivID].aivPoorCounter = 0;
    this->SEC_AIVS[aivID].aivSubType = 0;

    this->SEC_AIVS[aivID].aivPoorLimit_OR_AIC_buildInterval
        = MACRO_CALL_MEMBER(AICState_Func::getAIBuildInterval, DAT_AICState::ptr)(playerID);

    ++this->aivCount;
    return aivID;
}
return 0;
```

I read once that loop types are transformable into each other. If we have a counting variable, I think a classic for-loop was the most likely structure.
I did not manage to add a direct alias via ptr to the aiv spec to the loop to avoid the repeated access. Something in the handling of the ptr would then change. However, always accessing them via the run-variable seems to work and avoids the kinda awkward ptr creation.

My feeling would say that this could be very close to the original used syntax (excluding macro or constant stuff)... outside of the early continue maybe. But who knows...

gynt: ## Claude
**The refactoring is not possible while maintaining 100% match.**

Here's why: the repeated `this->SEC_AIVS[aivID]` pattern is not just a stylistic redundancy — it is the **exact codegen fingerprint** of the original MSVC 2005 compiler output. The assembly diff reveals what the original binary actually does:

- It uses `eax` for the loop pointer (computed via `lea eax, [ebx + 0x6d9c]` + `add eax, 0x6d98` per iteration).
- It accesses `[eax]`, `[eax + 4]`, `[eax + 8]`, etc. for each field — recomputing the address each time from `this + offset`.

When you introduce `AIVSpec& aiv = this->SEC_AIVS[aivID]`, MSVC 2005 at `/O2` keeps the reference address live in a **separate register** (`esi`), requiring an extra `push esi` / `pop esi` around the entire function — a different register allocation that breaks the match immediately.

The 100% match is only achievable precisely **because** the code spells out `this->SEC_AIVS[aivID]` each time, letting the optimizer emit the compact single-register loop that the original compiler produced. The repetition in source is the price of a perfect binary match.

src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
TheRedDaemon: Was done.

src/OpenSHC/AI/AIVState/setupAIVMetadata.cpp
TheRedDaemon: Was done.

src/OpenSHC/AI/AIVState.hpp
TheRedDaemon: Builds now.

TheRedDaemon (approved):

PR #64, opened 2026-05-12
reimplement: SHC_3BB0A8C1_0x004CB120 100%

gynt:
100% matching


src/OpenSHC/AI/AICState/getAIBuildInterval.cpp
```diff
@@ -0,0 +1,20 @@
...
+    // FUNCTION: STRONGHOLDCRUSADER 0x004cb120
+    int AICState::getAIBuildInterval(int playerID)
+
+    {
+        int _aiType;
+
+        _aiType = DAT_GameState::ptr->playerDataArray[playerID].aiType;
+        if (_aiType == 0) {
+            return _aiType;
+        }
```
TheRedDaemon: Would prefer:
- direct assign to `const int _aiType`
- Usage of the enum you already used in another PR
- `return 0` -> at least I think that is what is happening here.

TheRedDaemon (changes requested):

src/OpenSHC/AI/AICState/getAIBuildInterval.cpp
TheRedDaemon: Was done.

TheRedDaemon (dismissed):

TheRedDaemon (approved):

PR #65, opened 2026-05-13
reimplement: OpenSHC::Random::RNG 100%

gynt:
```raw
  OpenSHC::Random::RNG::setTimeBasedSeed (0x46a740) is 87.50% similar to the original
  OpenSHC::Random::RNG::populateRNG1040 (0x46a760) is 93.75% similar to the original
  OpenSHC::Random::RNG::nextRandomNumber2 (0x46a7d0) is 100.00% similar to the original
  OpenSHC::Random::RNG::nextRandomNumber1 (0x46a800) is 100.00% similar to the original
  OpenSHC::Random::RNG::Constructor_RNG (0x471810) is 100.00% similar to the original
```

The reimplementations for `setTimeBasedSeed` and `populateRNG1040 ` are not 100% because they use lib functions that are currently set to non-reimplemented by `REIMPLEMENTED_CRT`.


src/OpenSHC/Random/RNG/Constructor_RNG.cpp
```diff
@@ -0,0 +1,17 @@
+#include "OpenSHC/Random/RNG.func.hpp"
+#include "OpenSHC/Random/RNG.hpp"
+
+namespace OpenSHC {
+namespace Random {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x00471810
+    RNG* RNG::Constructor_RNG()
+
```
TheRedDaemon: This newline is valid according to our linter settings?
Well, ok. 😅

src/OpenSHC/Random/RNG/nextRandomNumber1.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x0046a800
+    void RNG::nextRandomNumber1()
+    {
+        this->currentNumber1 = this->randomNumbers[this->index1];
+        this->index1 = this->index1 + 1;
+        if (20000 <= this->index1) {
+            this->index1 = 0;
+        }
+        return;
```
TheRedDaemon: I do not think we should use an explicit final return in a void function.

src/OpenSHC/Random/RNG/populateRNG1040.cpp
```diff
@@ -0,0 +1,36 @@
...
+        this->index2 = 0;
+        this->index1 = 0;
+        short* _pRandomNumber = &this->randomNumbers[0];
+        int n = 20000;
+        do {
+            int random = MACRO_CALL(OpenSHC::OS_Func::_rand)();
+            *_pRandomNumber = (short)random;
+            _pRandomNumber += 1;
+            n -= 1;
+        } while (n != 0);
```
TheRedDaemon: This is unnerving... I would think running over the whole `randomNumbers` array with a loop assigning numbers would be sensical, but this runs positiv to 0, but 0 index to 19999...why...

I assume making this an up-counting for loop will not create the same effect, hm?
Can you try something like this (although I assume it is likely this does not produce the same code):
```
for (int i = 0; i < 20000; ++i)
{
    this->randomNumbers[i] =  static_cast<short>(MACRO_CALL(OpenSHC::OS_Func::_rand)());
}
```

status/addresses-SHC-3BB0A8C1.txt
```diff
@@ -10168,10 +10168,10 @@ SHC_3BB0A8C1_0x0046A4C4 | 0.0% | Pending
 SHC_3BB0A8C1_0x0046A4C8 | 0.0% | Pending
```
TheRedDaemon, line 10168: Do you count the one that work if the Lib functions are active as 100%?

TheRedDaemon (changes requested):

src/OpenSHC/Random/RNG/Constructor_RNG.cpp
gynt: I don't like it either hahaha, shall we update the clang-formatter? :)

src/OpenSHC/Random/RNG/nextRandomNumber1.cpp
gynt: fair!

status/addresses-SHC-3BB0A8C1.txt
gynt: Yes! I was not sure what to call them. "Reimplemented (non-call matching)"

src/OpenSHC/Random/RNG/Constructor_RNG.cpp
TheRedDaemon: If you find a fitting setting, feel free.
It is not horrible if some of this slips through, but better earlier than later.

status/addresses-SHC-3BB0A8C1.txt
TheRedDaemon: I have a similar case in #56. It really depends a bit on how we handle the resolvers. I wrote you on Discord that the partial tests might get harder the more resolvers are active.

If we assume we keep the resolver on false in general, a reimplementation would be 100% if it matches 100% in case the called resolvers are active. In would not make sense otherwise and would not need a special mark, since the non-100% are artifical through the resolvers.

src/OpenSHC/Random/RNG/Constructor_RNG.cpp
TheRedDaemon: So, short research indicates that there is no specific setting to remove this line break. I think than it is okayish, but if these things happen, try to remove them manually.

src/OpenSHC/Random/RNG/populateRNG1040.cpp
TheRedDaemon: So, this one is interesting:
It actually works this way and generates the fitting bytecode. Having an inverse counter just seems to be something the compiler decided, so you can use this simpler loop.

I checked how one would get the size of an C array also, but the only thing I found was `sizeof(array) / sizeof(array[0])`, so one could make this:
```cpp
for (int i = 0; i < sizeof(this->randomNumbers) / sizeof(this->randomNumbers[0]); i++) {
    this->randomNumbers[i] = MACRO_CALL(OpenSHC::OS_Func::_rand)();
}
```
One could also make the size of randoms a constant or something, to use in different positions. But I am not sure if this kind of "binding" is something we want to do at this point, since it makes an overreach, leaving the current cpp file.

But it makes my suspicion likelier, that structures like `short* _pRandomNumber = &this->randomNumbers[0];` outside of a loop are in general an indicator for a for loop with array access.

status/addresses-SHC-3BB0A8C1.txt
gynt: Sounds good to me!

src/OpenSHC/Random/RNG/populateRNG1040.cpp
gynt: This doesn't work always `sizeof(this->randomNumbers) / sizeof(this->randomNumbers[0]` because it creates an unsigned leading to asm `jb` instead of `jl` (for signed numbers). So pretty sure they used the number literally.

TheRedDaemon (changes requested):
Can you also remove the other explicit non-early `returns`?

TheRedDaemon (approved):

PR #68, opened 2026-05-14
reimplement: SHC_3BB0A8C1_0x00424700 100%

TheRedDaemon:
I will include the resolver switch and/or adding to the complete compile list depending on our decision.


src/OpenSHC/Audio/mss/SoundSystem/meth_0x424700.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "OpenSHC/Audio/MSS/enums/SHC_SoundStream.hpp"
+
+namespace OpenSHC {
+namespace Audio {
+    namespace MSS {
+
+        // FUNCTION: STRONGHOLDCRUSADER 0x00424700
+        void SoundSystem::meth_0x424700()
+        {
+            const UnkSoundFlagsAndLoopCount unkSoundFlagsAndLoopCount = {};
```
gynt: Do you prefer `const <type>`? I prefer `<type> const` because it is more consistent when pointers get involved `char const * const`.

gynt (changes requested):

src/OpenSHC/Audio/mss/SoundSystem/meth_0x424700.cpp
TheRedDaemon: I am basically Java trained, where the qualifiers are always before the type, JavaScript the same, so that is just natural behaviour.
But I think I see you point, considering that C++ types are ~a mess~ complicated.

I changed the notation here and added a merge request for qualifier formatting (#69),

gynt (approved):

PR #72, opened 2026-05-15
reimplement: SHC_3BB0A8C1_0x00479AB0 100%

TheRedDaemon:
Draft, since there are two problems at the moment:
- "inaddr.h" was not found. Might be an internal header?
-  IDirectPlay4A has a recursive definition going on it seems. I broke it with removing the one `using` and forward declaring it. The question here however would be: Are these structures relevant for us, or can we just use "dplay.h"?


TheRedDaemon:
Hi @gynt, I still have the issue with the `inaddr.h` header:
```
src\OpenSHC/Synchrony/GameSynchronyState.hpp(33) : fatal error C1083: Cannot open include file: 'inaddr.h': No such file or directory
```
Can I just leave the removal of this include in this MR and you adapt the export, or should we do something else?

gynt:
Resolved in a1971f306e193e55ed48af360f99d65dd7b37f08

gynt (approved):

PR #81, opened 2026-05-20
reimplement: SHC_3BB0A8C1_0x00479E60 100%

TheRedDaemon:


src/OpenSHC/Audio/mss/SoundSystem/setVolumeUnk.cpp
```diff
@@ -0,0 +1,39 @@
...
+        {
+            if (!this->waveOutOpenUnk_0x8) {
+                return;
+            }
+
+            if (streamIndex == -1) {
+                this->sampleVolume = volume;
+                for (int i = 0; i < 31; ++i) {
+                    // TODO: Most structures are either 32 or 1 and 31. We need to unify these.
+                    //   Either everything else is of or sample should take in the int value before.
```
gynt: I don't understand these comments. If you do then no problem. I just want to avoid comments from ghidra appearing automatically in this codebase

gynt (dismissed):

src/OpenSHC/Audio/mss/SoundSystem/setVolumeUnk.cpp
TheRedDaemon: I tried to improve the comment.
The comment is from me and just notes something I noticed during reimplementation.
It is a general note should someone ever get to refactoring or something like this.

gynt (approved):

PR #83, opened 2026-05-22
refactor: directplay and dplobby redirect to dplay.h and dplobby.h, include dplobby.h

gynt:
I think we forgot dplobby.h. I added it to dependencies from the same source as you took the other ones from. 

The only objects left in OpenSHC/DirectPlay/ are helper enums. All the other header-imitating objects are in Ghidra in subdir dplay/ and dplobby/ and excluded from ghidra export into this repo as we get those from the two header files instead.

Also points to proper source for IN_ADDR, which is WinSock.h instead of inaddr.h

Resolves issues in #72


src/OpenSHC/Synchrony/GameSynchronyState.hpp
```diff
@@ -27,8 +24,11 @@
 #include "OpenSHC/UI/Enums/MenuModalTypeInt.hpp"
 #include "OpenSHC/WindowsHelper/Enums/BOOLEnum.hpp"
 #include "WinDef.h"
+#include "WinSock.h"
 #include "basetsd.h"
 #include "crtdefs.h"
+#include "dplay.h"
+#include "dplobby.h"
 #include "guiddef.h"
 #include "inaddr.h"
```
TheRedDaemon, line 33: Is it ok that this is still included?

TheRedDaemon (changes requested):

src/OpenSHC/Synchrony/GameSynchronyState.hpp
gynt: What do you mean with "still included"? I just added them in place of "#include OpenSHC/DirectDraw/*.hpp". The GameSynchronyState.hpp file uses those structures.

Or do you mean it is unnecessary to include them as they are a dependency?

TheRedDaemon: When I tested it in #72, "inaddr.h" could not be found. The file itself was unknown.
Maybe my question should therefore rather be: If you include `src/OpenSHC/Synchrony/GameSynchronyState.hpp` somewhere, does it compile, or is still an error thrown?

gynt: It compiles! I tested it. Inaddr header references should have been removed by this pr.

TheRedDaemon (approved):

PR #87, opened 2026-05-24
reimplement: SHC_3BB0A8C1_0x00471830 and support functions 100%

### LINKED ISSUE #91: [PROJECT] Check methods of "LowLevelMemory" regarding possible older compilers
After trying another method again, the patterns still barely match.
My last try was 0x0046AE50 (copyStringUntilFirstDot).
It could still be possible they used very unusual loop forms or techniques, or simply structures I do not think of. But the only way I found so far for re-creating the match was using inline assembly for the whole function body, which properly formed the prolog and epilog to what is was.

---

We have to consider that older compilers or libraries may be involved. The may have used an older static support library in the exe.
So part of this issue would be testing one or two of the smaller "LowLevelMemory" functions with older compilers and checking if one gets better results.

---

An attempt with [MSVC7.00](https://github.com/archaic-msvc/msvc700) failed completely, since only 7.10 somewhat fixed the template support.

Therefore, tests need to be performed on a very small scale, maybe the resolvers can be excluded for these test from the header, to still keep the ENV.
Also, note that MSVC1400 is also mentioned in the CMakeLists.txt.

TheRedDaemon:
The weird patterns in the copy functions made me use assembly in the end.
Maybe that is the origin, maybe it is not. I found no other way to recreate this pattern.
Is heavily commented.

Also, draft: There needs to be a type change in `LowLevelMemory.hpp`.

Closes #91 .

### DIFF
--- a/src/OpenSHC/IO/LowLevelMemory/copyData+.cpp
+++ b/src/OpenSHC/IO/LowLevelMemory/copyData+.cpp
@@ -3,9 +3,9 @@
 namespace OpenSHC {
 namespace IO {
 
-    // NOTE: The calls to the copyData... functions are optimized, which means the compiler knew the content of these
-    //   functions. The most likely thing is, that these functions shared a file. It does therefore only work if the
-    //   resolvers are active.
+    // NOTE: The calls to the copyData... functions are optimized in "copyData", which means the compiler knew the
+    //   content of these functions. The most likely thing is that these functions shared a file. It does therefore only
+    //   work if the resolvers are active.
 
     // FUNCTION: STRONGHOLDCRUSADER 0x00471830
     void LowLevelMemory::copyData(size_t size, void* src, void* destination)
@@ -24,8 +24,8 @@ namespace IO {
     //   to not use the stack and use a more optimized loop. Regardless what the original reason was, no other way was
     //   found so far aside of using assembly, which provokes the strange "store in stack then load" pattern.
     //
-    //   Since these are low level support functions anyway, I would consider this good enough. They can be visited if
-    //   anything else is done to be rejudged.
+    //   The general sentiment expressed by these functions is added as a comment before the assembly code. Note, that
+    //   actually using these would not result in the same binary.
 
     // FUNCTION: STRONGHOLDCRUSADER 0x0046ABA0
     void LowLevelMemory::copyData0x100()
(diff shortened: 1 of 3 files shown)


TheRedDaemon:
Added logic comments, however, still needs change in class definition of `LowLevelMemory.hpp`, either confirmed to have happend in Ghidra, or via commit to main.

gynt (changes requested):
Can you change the name of file `copyData+.hpp` to something without the `+`?
Or split them up again in separate files or does that ruin the match?
I would like to stick to the philosophy of the flie location being locatable just by function name and namespace.

TheRedDaemon:
I can simplify it again to `copyData.hpp`, but they need to be in one file (under the assumption that Global Optimization was not used by them). Here my comment:
```
    // NOTE: The calls to the copyData... functions are optimized in "copyData", which means the compiler knew the
    //   content of these functions. The most likely thing is that these functions shared a file. It does therefore only
    //   work if the resolvers are active.
```

This may happen to us in other cases, too. At least in any moment Ghidra has issues to reuse the ECX, we have to assume they shared a file, and we have to put them together for the perfect match.

gynt:
I synced the change in .hpp in a1971f306e193e55ed48af360f99d65dd7b37f08

TheRedDaemon:
Still requires a decision regarding file naming: Should I just call it "copyData"?

gynt:
So the current decision comes down to the following:

**Option A**
Have different files for each function, which makes it (for DLLs) not possible for the linker/compiler to improve `copyData` to not inject multiple `mov ecx, ...` statements for each sub call. Therefore, `copyData` should be inline assembly as well.

This keeps the binding from function namespace to filename intact.

**Option B**
Put the functions in the same file, allowing the compiler/linker to optimize `copyData`. This is more true to the original.

This ruins the binding from function namespace to filename. We could also "encode" (or make it machine predictable), by renaming the sub functions to: `copyData_0x001` etc. Then the `_` denotes the sub-ness of the function. We could also say these were private functions of the class, which would call for `_copyData_0x001`.
Then we keep some predictability: function copyData_blabla is to be found in copyData.cpp

What do you think of my B naming proposal?

TheRedDaemon:
I think I prefer B, since it tries to be more matching.
Using an explicit "sub"-naming like `<file>-<sub-function>` also feels future proof. We might encounter this structure more often and with this we stay flexible in case essentially "unrelated" function were implemented in the same file.
In this case here, one function might stay `copyData`, but we could end up in a situation where we only have sub-functions in a file.

So B. Do you want to lead with the re-naming?

gynt:
Yes I will rename this upstream. Then we merge/rebase main into this branch and merge this PR.

gynt:
Renaming upstream has been done! Do you wanto to merge/rebase?

gynt (approved):

PR #94, opened 2026-05-28
reimplement: SHC_3BB0A8C1_0x00479FC0 100%

TheRedDaemon:


gynt (approved):

gynt:
The progress.html shows a diff entry that I find a little suspicious: might the original code have looked very slightly differently? It is the `lea ebx. [ebx]` I am talking about. Why does our reimplementation include that? @TheRedDaemon 

image

PR #95, opened 2026-05-30
reimplement: SHC_3BB0A8C1_0x0047A130 100%

TheRedDaemon:
Requires changes to `AttackInfo`: Two new field sizes set.

This one showed the shortcomings of the struct resolver:  
Unlike the function resolver, which seems to be pretty good at behaving like a normal function, the struct resolver deeply effects the way the compiler handles field accesses. It required activating almost all of them to see the opcodes being produced like in the original. There is a clear difference between having the global or just a reference, no matter how stable. We always have to keep this in mind while testing around.

Also:
- The `private` functions should be made public. Pragmatic reason: We can not judge visibility and we can not instantiate classes with constructor if they are private.
- I found two struct resolvers for SHC_3BB0A8C1_0x01A279C0: In the class definition and as global. Is this intentional?

### DIFF
--- a/src/OpenSHC/Audio/mss/SoundSystem/mapLoadingAndLaunchGameRelated1.cpp
+++ b/src/OpenSHC/Audio/mss/SoundSystem/mapLoadingAndLaunchGameRelated1.cpp
@@ -23,9 +23,9 @@ namespace Audio {
             DAT_GameCore::ptr->battleLevel = 0;
             DAT_GameCore::ptr->battleLevel2 = 0;
             DAT_GameCore::ptr->field36_0x90 = 0;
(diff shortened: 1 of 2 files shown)


gynt:
The class definition struct resolver should be removed.
For the record, which class were you trying to instantiate but couldn't because of private constructors?

gynt:
I resolved the AttackInfo.hpp thing now in a1971f306e193e55ed48af360f99d65dd7b37f08

TheRedDaemon:
@gynt I think it was "GameCore" and "TroopValueState", but I am not 100% sure anymore. Should have been every class that has a constructor declaration that was private.

gynt (approved):

PR #97, opened 2026-05-31
reimplement: SHC_3BB0A8C1_0x0047A220 100%

TheRedDaemon:


src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
```diff
@@ -0,0 +1,39 @@
...
+                }
+            }
+
+            if (soundID == this->currentSoundID_0x3278) {
+                return;
+            }
+            this->currentSoundID_0x3278 = soundID;
+            int const fileVolume = MACRO_CALL_MEMBER(SFX::SFXState_Func::getSoundVolumeForFilename, DAT_SFXState::ptr)(
+                DAT_SFXDefinedData::ptr->DAT_SFX_Pointers[soundID].musicFile);
+            this->currentSoundIDVolumeUnk_0x327c = (fileVolume * soundMultiplier) / 0x7f;
```
gynt, line 36: What should we do with magic values like `0x7f` and `19` and `46`? Perhaps the best way to go about this is to document this (perhaps with enum constants) in a different PR later on and focus on the matching for now?

gynt (dismissed):

src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
TheRedDaemon: Naming them would mean understanding them, which I am not sure of. 🙂 
Assuming the `SHCDE` enums still fit the old game, then for some reason, if the sound ID is `MUSIC_TUNE_NARR1`, it checks if the mission number is between 1 and 20 and if yes, applies 46 to the mission number, which at its lowest would cause the soundID to be `MUSIC_PRE_MISSION1`. There is space for 20 enums in theory before `MUSIC_AFTERMATH` is the next enum, even if only 4 are used.
So assuming `MUSIC_TUNE_NARR1` is not mislabeled, it has become an indicator to check if one of the mission sounds should play, weirdly. In theory, the function could just use the actual mission ids, but who knows how this ended up here.
So in theory, these numbers are less enums and more enum movers... I could add a comment?
But even then, I am also more inclined to see further analysis as a bonus, something to be done out of interest or "after" reimplementation. Our process makes it a hassle otherwise, since every definition change needs another request to adapt the Ghidra stuff.

0x7f is 127 btw., the upper limit for the sample volume in MSS, assuming the comment I added to mss32 was properly copied.

gynt: No it is fine, let's do this later.
Add what you have as a comment for now!

gynt (approved):

PR #108, opened 2026-06-13
reimplement: SHC_3BB0A8C1_0x0047A790 100%

TheRedDaemon:
Requires changes in `UnkSoundFlagsAndLoopCount.hpp` (not sure if ok, since separated from Ghidra? Still has the autogenerated disclaimer) and `SoundEffectsHelperData1.hpp`.
I also wait until I can test with the reccmp struct comments, since there were many.


gynt:
Partly addressed in 7e1149e1
Rest is addressed in c2e2e939

gynt (approved):

PR #113, opened 2026-06-15
reimplement: SHC_3BB0A8C1_0x0047AB10

TheRedDaemon:
~Hey @gynt, want to waste some credits/tokens?~

~Spend the good half of a whole day on this to no success. Combines goto and complex conditionals.~

~The current file contains a commented out version of my furthest try:~
~- GOTOs removed~
~- Loops that enroll~
~- Stack start worked~

~Issues however:~
~- Fails to recreate switch~
~- The value `volumeLevel` (I really think the name is wrong, at least based on the logic in this function) is placed into a register instead of being written directly and loaded, likely because under this condition the compiler can proof that it can cache it. This messes up the whole code. Doing something that forces the store can increase the matching value by 30 to 40 percent while being likely wrong logic-wise.~

~The other version (and currently reccmp commented) is almost the prepared decomp from Ghidra. I just flipped some if-cases around that were clear. MSVC seems to produce inverted if logic in a lot of cases compared to what Ghirda assumes.~

~Other then this the commit has the instances activated to remove issues with the struct access in the first place. Should someone solve this, then they need to be deactivated again.~

~Need support with this.~

### DIFF
--- a/src/OpenSHC/Audio/mss/SoundSystem/playRandomMusic02.cpp
+++ b/src/OpenSHC/Audio/mss/SoundSystem/playRandomMusic02.cpp
@@ -1,0 +1,187 @@
+#include "../SoundSystem.func.hpp"
+
+#include "OpenSHC/Map/Units/TroopValueState.func.hpp"
+#include "OpenSHC/Random/RNG.func.hpp"
+
+#include "OpenSHC/Globals/DAT_GameSynchronyState.hpp"
+#include "OpenSHC/Globals/DAT_SoundEffectsHelperData1.hpp"
+#include "OpenSHC/Globals/DAT_TroopValueState.hpp"
+#include "OpenSHC/Globals/SEC_RNG.hpp"
+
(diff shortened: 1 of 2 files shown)


TheRedDaemon:
No further help needed.
Found the main reason:
One if else case was a switch originally, causing completely different register usage.
Will try to match 100%.

TheRedDaemon:
There is still a diff with a type value. It was set to `uint`.
Any reason for this change? I think a condition was handled in a way that indicated signed int in this function.

gynt:
> There is still a diff with a type value. It was set to `uint`. Any reason for this change? I think a condition was handled in a way that indicated signed int in this function.

I went: let's remove all the undefined stuff from that struct and be ahead of TRD! Then I realized, let's undo that because it might mess with stuff if I do anything wrong. I guess I forgot to undo that one. Sorry!

EDIT: fixed in main

gynt (approved):

PR #114, opened 2026-06-15
[SoundSystem] fix files using the enum parameters

TheRedDaemon:
Tested, produces same match level.

### DIFF
--- a/src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
+++ b/src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
@@ -12,26 +12,28 @@ namespace Audio {
     namespace MSS {
 
         // FUNCTION: STRONGHOLDCRUSADER 0x0047A220
-        void SoundSystem::setupVolumeAndSoundIDWithMultiplier(eMusicIDsInt soundID, int soundMultiplier)
+        void SoundSystem::setupVolumeAndSoundIDWithMultiplier(eMusicIDs soundID, int soundMultiplier)
         {
+            int soundIDTemp = (int)soundID;
+
             // Indicator to handle as mission sound (?)
-            if (soundID == DE::SHCDE::MUSIC_TUNE_NARR1) {
+            if (soundIDTemp == DE::SHCDE::MUSIC_TUNE_NARR1) {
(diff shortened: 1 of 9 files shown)


src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
```diff
@@ -12,26 +12,28 @@ namespace Audio {
     namespace MSS {
 
         // FUNCTION: STRONGHOLDCRUSADER 0x0047A220
-        void SoundSystem::setupVolumeAndSoundIDWithMultiplier(eMusicIDsInt soundID, int soundMultiplier)
+        void SoundSystem::setupVolumeAndSoundIDWithMultiplier(eMusicIDs soundID, int soundMultiplier)
         {
+            int soundIDTemp = (int)soundID;
```
gynt: Is this required for matching?

gynt (approved):

src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
TheRedDaemon: It is easier compared to always casting the value on usage. And it matches like before.

It is still very likely these were only ever ints originally, but since we are using the enums to increase readability of the code, we have to deal with casting. I think that is acceptable.

gynt: They probably used macro's then to keep track of them. But setting those in Ghidra is a pain (Equate is the term) and uses Enum's internally. So let's keep it the way it is.

gynt: Can you change soundIDTemp to `soundIDInt` ? At least that makes it clear enough. Temp sounds silly if it is used over and over.

gynt (changes requested):

src/OpenSHC/Audio/mss/SoundSystem/setupVolumeAndSoundIDWithMultiplier.cpp
TheRedDaemon: Renamed

gynt (approved):

PR #116, opened 2026-06-16
[EXPERIMENT/STRUCT_RESOLVER] Handle the not implemented struct resolver case via asm file symbols

### LINKED ISSUE #115: [STRUCT_RESOLVER] Rework to using MASM linker trick instead of pointer
The current struct resolver does something similar to the function resolver and provides a ptr in case it is not active.
However, this causes issues. Unlike the functions, where inlining the function ptr on usage does not cause any difference between the reference to the game and the reimplementation, the code changes in case of variables.

However, thanks to chatbots, there is another workaround using MASM files.

---

Assuming we have a variable of type `OpenSHC::Audio::SoundEffectsHelperData1` named `test` in global scope:
```
extern OpenSHC::Audio::SoundEffectsHelperData1 test;
```

I can not create an asm file, for example `test.asm`, looking like this:

```asm
.386
.model flat

PUBLIC ?test@@3USoundEffectsHelperData1@Audio@OpenSHC@@A

_TEXT SEGMENT
?test@@3USoundEffectsHelperData1@Audio@OpenSHC@@A EQU 00DF37F0h
_TEXT ENDS

END
```

If I compile this with this:
```batch
ml /c /coff test.asm
```
and add the resulting obj to the dll target in CMake, then the build succeeds, since the linke just uses the address given by the object file.

---

This might be enough of a win to rework the struct resolver.
The function resolver is likely fine as it is, since I am not sure if this could provide any way to just use the function definition and still use the proxy. Likely not.

What still need to be researched is how to create the symbols and how to combine it with the main resolver feature of having a code internal toggle.
It might be possible to crete the externs with externs C to better control the naming.

TheRedDaemon:
Closes #115

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ function(file_dependent_read_list FILE LIST)
     set(${LIST} "${TEMP_LIST}" PARENT_SCOPE)
 endfunction()
 
-project ("OpenSHC")
+project ("OpenSHC" LANGUAGES C CXX ASM_MASM)
 
 # Name of exe and dll should be the same
 set(OPEN_SHC_NAME "OpenSHC")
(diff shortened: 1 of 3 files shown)


TheRedDaemon:
Superceded from #125

PR #119, opened 2026-06-20
reimplement: SHC_3BB0A8C1_0x0047B060 100% + MSS32 fix

TheRedDaemon:
Removed the open flags, since Windows provides MACROs for them.

~This function does not seem to be compatible with the bitfields, but even if I use an int, the last "loopCount" usage is still optimized to an 16bit mov instead of the "and 0xffff".~
~Yet still, even if I assume they used a plain int, I would still be confused regarding the bit compares, since that is horrible to use. Maybe they used helper macros? I managed to make the other cases work so far, increasing my confusion.~
While it seems they are not compatible with bit fields, certain changes to the structure of the code cause specific parts to be properly handled like in the original. Even the current case contains one such compare.

In any case, the function has some non-fitting register usage even if the bitfield usage is transformed, so something is wrong anyway.
The logic itself should be sound.

---

EDIT:
Fixed issues, two things were discovered:
- The typing of one Mss32 function was wrong. I made a mistake when creating our simple header. Confirmed this with the real header.
- It was not possible to directly use `0` for the music stream, instead I even needed to negate the current sampleIndex in one case. I can not confirm if this is due to the original being this way, or if the reimpl did something different, but once again something to keep in mind in some cases.


gynt (approved):

gynt (dismissed):

gynt (dismissed):

gynt (approved):

PR #120, opened 2026-06-20
reimplement: SHC_3BB0A8C1_0x0047b250 100%

TheRedDaemon:
When using the bitfield, the compiler uses a nulled EAX to send the 0 parameters to "playSoundStreamUnk".
I only managed to get the two 0 pushes from the original to work when using an "int" (might also be unsigned int).

Once again, everything becomes inconclusive. The bitfields are clearly capable to producing the code of the exe under certain circumstances, including the combined bit compares.
But this calls this somewhat into question, since the function is rather small with not a lot of room, outside of maybe wrong types here and there.

I am wondering if a switch to and `int` or `unsigned int` would be more stable at this point, but the usages really suggest helper structures to me. But even if they used a bitfield caster or helper macros, I have no idea where I should store them...

The logic itself is equal.


gynt (dismissed):
bitfield `UnkSoundFlagsAndLoopCount` has become `typedef unsigned int SoundFlagsAndLoopCount`

gynt (approved):

PR #124, opened 2026-06-20
reimplement: SHC_3BB0A8C1_0x0047B590 100%

TheRedDaemon:
Yep, once again full circle:
I suggest to remove the `UnkSoundFlagsAndLoopCount ` as value type completely and instead try to use bit operations.
This function here basically consists out of nothing and assuming there are no weird optimizations related to knowing the called functions, only plain values produce the required order.
The only other way would be compiler differences, but these have lesser probability.

That being said, I would prefer if we could use some constants/macros for this with specific names. These here in the PR are only examples. I am open for better suggestions.


src/OpenSHC/Audio/mss/UnkSoundFlagsAndLoopCount.hpp
```diff
@@ -11,21 +11,12 @@
...
-        // SIZE: 0x00000004
         typedef struct UnkSoundFlagsAndLoopCount {
-
-            unsigned int loopCount : 16;
-            int reserved : 13;
-            int unknownFlag1 : 1;
-            int unknownFlag2 : 1;
-            int uninterruptable : 1;
-
+            static unsigned int const LOOP_COUNT_FIELD = 0x0000FFFF;
```
gynt, line 25: What about dropping `UnkSoundFlagsAndLoopCount ` altogether in favor of an `int`, and move these to an `enum`. So then in function bodies you can rely on the enum entries to do the manipulations.

```cpp
int loopCountFieldPart = value & LOOP_COUNT_FIELD;
int uninterruptible = value | UNINTERRUPTABLE;
```

TheRedDaemon: I was thinking about an enum, but I am not really sure if this properly conveys the mixture of flags and a value. At the moment I am also scared that the enum might have yet other effects, even if this seems unlikely so far.
But we could try? Even if it turns out it is once again a problem, we could then just change it to constants.

I will try to finish the methods of SoundSystem and then experiment with the remaining non-fitting functions.
Types is also some of the last things I could try.

gynt: Okay unsigned int const for now if that is more likely right now to achieve 100% results.

gynt (dismissed):
bitfield `UnkSoundFlagsAndLoopCount` has become `typedef unsigned int SoundFlagsAndLoopCount`

gynt (approved):

PR #125, opened 2026-06-20
Implement extern symbols with name mangling

gynt:
Achieved through ghidra-skink library update with a custom build MSVC1400 name mangling python library.
Covers all global defined data in the current project.
Tested by including all OpenSHC/Globals/*.hpp files in a single cpp file.

Renamed `game.asm` to `data.asm` but I am open to suggestions. I suppose if we do more with asm in the feature we will add more asm files so I thought `game.asm` was a bit too generic of a name.

This PR builds upon #116


TheRedDaemon (approved):
Awesome.
I will merge this and close my PR and do some clean ups on main directly.
This should really help us getting proper matching values when it comes to the functions.

PR #126, opened 2026-06-21
reimplement: SHC_3BB0A8C1_0x0047B890 and SHC_3BB0A8C1_0x0047a340 both 100%

TheRedDaemon:
@gynt:
I managed to match both functions (SHC_3BB0A8C1_0x0047B890 was something...).
Still requires some type adjustments.
Also, there are two FUNCTION_RESOLVERS that have address issues, see the diff of the commit.


src/OpenSHC/Audio/SFX/SFXState.func.hpp
```diff
@@ -80,7 +80,7 @@ namespace Audio {
             PlaySpeechSFX;
 
             MACRO_FUNCTION_RESOLVER(
-                void (SFXState::*)(char*), false, Address::SHC_3BB0A8C1_0x0044A460, &SFXState::playWAVSFX)
+                void (SFXState::*)(char const*), false, Address::SHC_3BB0A8C1_0x0044A460, &SFXState::playWAVSFX)
```
gynt, line 83: Are these const modifiers a habit or are they required for the correct byte code?

TheRedDaemon: String literals are `char const*`. I can not pass them to the function without it being definied as const.
In turn, at least any parameter in the game that takes a string literal should be `char const*`.

gynt: Ah right that makes sense. It is just that Ghidra doesn't support `const` modifiers. So I will solve this with sugar in the exporter: in Ghidra database declare a typedef CHAR_CONST (that in Ghidra itself just points to char), then add a rewrite rule in my exporter to convert `CHAR_CONST` in `char const` upon export for function arguments.

gynt:
> Also, there are two FUNCTION_RESOLVERS that have address issues, see the diff of the commit.

What issue do you mean specifically?

TheRedDaemon:
These two addresses do not exist in the `Address`-Enum:
- SHC_3BB0A8C1_0x004719E0
- SHC_3BB0A8C1_0x0046B340

So this would fail to compile since `GameCore.func.hpp` is used.

gynt:
That is strange why just these two functions are exported differently. I don't see anything in the sarif file that could cause this behavior.

hahaha.... found it. I am ignoring symbols that start with "switch" and I accidentally applied that logic to functions too...

gynt (approved):

PR #128, opened 2026-06-25
reimplement: SHC_3BB0A8C1_0x0047C400 100%

TheRedDaemon:
This one has a type change with rather big implications:
Every usage of `sample_0x190` needs to be adapted to `sample_0x18C`, which based on the contained loop is like many other structures 32 ints.

Outside of this, there is the usage of two non-member funcs here. Where should the files of these go, actually?


gynt:
For example `RegisterMilesAILShutdown` should go in `src/OpenSHC/Audio/MSS/RegisterMilesAILShutdown.cpp`. So essentially split the namespace via `::` and append `.cpp` :)

gynt:
fixed upstream

gynt (approved):

PR #142, opened 2026-07-03
refactor: name unnamed functions with a best guess

gynt:
This is a big one. I renamed all unnamed functions and clarified some others. So I would like to have a review from others. Are these function names clear?

### DIFF
--- a/src/OpenSHC/AI/AICState.func.hpp
+++ b/src/OpenSHC/AI/AICState.func.hpp
@@ -587,8 +587,8 @@ namespace AI {
         decideOnTribeAttackLocation;
 
         MACRO_FUNCTION_RESOLVER(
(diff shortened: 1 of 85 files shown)


TheRedDaemon:
What do you expect from the review?
Should I check Ghidra and try to confirm your reasoning? Or is is just about if the names are nice?

gynt:
I wasn't that sure. It felt too big to just push to main without a PR :)

It has consequences for the work that has been done to SoundSystem because of renames. Should a rename of the cpp files be included in this PR? Should I hold back the one that is currently open in the PR?

TheRedDaemon:
I think if they are already part of main and you rename something, you definitly need to pull the cpp files with you (rename the file and the function inside).
Functions that you know are being worked on however is not something you need to take into consideration outside of a comment on the PR for now. 
That being said, these things might slip through should we ever become more, since we have no build validation yet. So it is something to keep in mind.

PR #145, opened 2026-07-07
reimplement: SHC_3BB0A8C1_0x0057E160 100%

TheRedDaemon:
Changed return and type based on on a helper function I found inside the header.
Also changed `atexit` parameter based on Win32 docs.


src/OpenSHC/OS/OS.cpp
```diff
@@ -42,7 +42,7 @@ namespace OS {
     int __stricmp(char* _Str1, char* _Str2) { return _stricmp(_Str1, _Str2); }
 
     // STUB: STRONGHOLDCRUSADER 0x0057FA62
-    int _atexit(void* param_1) { return atexit(reinterpret_cast<void(__cdecl*)()>(param_1)); }
+    int _atexit(void(__cdecl* func)()) { return atexit(func); }
```
gynt: ~~Does `_onexit_t` as a param type work? `int _atexit(_onexit_t func)`. That type already exists in Ghidra too.~~

gynt:
See fcc95ee9

gynt (approved):

PR #148, opened 2026-07-08
reimplement: SHC_3BB0A8C1_0x0044B840 84% and SHC_3BB0A8C1_0x0044AAB0 100%

TheRedDaemon:
Not much fun with this one.
The code has a lot of freedom due to a lot of math.

Worse (?) though is an issue with reccmp: There are array accesses that reccmp does not understand, it seems, which match only if I produce the original address, others however want the reimplementation to match. This makes trying to reason about the parts that seem to fit and those that do not hard.
I send a message to their discord.

I would halt this for now, since it is in no state to be merged. I think there is still a lot to try, but I spend a good amount of time trying and failing to match another function completely recently and in this other case reccmp **was** behaving at least.
We should return later in the hope the tool can improve, unless someone wants a challenge.

---

@gynt: I received help from the reccmp dev. The issue is the space between the type and the array brackets. The pdb data has a space there, so our annotations need to replicated this:
`int[...]` -> `int [...]`

---

EDIT: As expected, something still messes with the registers. Lost all patience on the other bigger functions.
Checked the logic, asked the AI, it is really just register choice for the most part and this thing has plenty of freedom.
There is also stuff like "DAT_DestroyedBuildingsCountHistory" and "DAT_DestroyedBuildingsCount", which could also be placed in a way that might suggest one array of 5, but replacing the type with this did not yield better results.

@gynt unless you want to try something, I would call it for this function.
Still needs the helper function adjustments, though.


gynt:
@TheRedDaemon Like so?
```diff
  // GLOBAL: STRONGHOLDCRUSADER 0x00B95B40
-// StructResolver::Instance<int[9],12147520>::instance
+// StructResolver::Instance<int [9],12147520>::instance
```

I tried to test but can't decide if this is wrong code or wrong comment notation by me

image

```cpp
// GLOBAL: STRONGHOLDCRUSADER 0x00B98630
// StructResolver::Instance<int [10],12158512>::instance

namespace OpenSHC {

MACRO_STRUCT_RESOLVER(int[10], false, Address::SHC_3BB0A8C1_0x00B98630) DAT_CurrentPlayerUnitLossesHistory;
} // namespace OpenSHC

MACRO_STRUCT_INSTANCE(Address::SHC_3BB0A8C1_0x00B98630);
```

gynt:
See fcc95ee9

gynt:
See 2d8606cc

gynt (approved):

PR #151, opened 2026-07-10
reimplement: SHC_3BB0A8C1_0x00449B30 100%

TheRedDaemon:
Two things:
- That is the second time a pointer was a `pointer *`. There might be an issue with the export?
-  Also, I found another restriction with the annotations for the globals: Alias for types are not allowed. A `unknown4` is an `unsigned int` for the compiler.


src/OpenSHC/Globals/DAT_LoadingBarProgress.hpp
```diff
@@ -9,7 +9,7 @@
 #pragma once
 
 // GLOBAL: STRONGHOLDCRUSADER 0x00D7D2D0
-// StructResolver::Instance<undefined4,14144208>::instance
+// StructResolver::Instance<unsigned int,14144208>::instance
```
gynt, line 12: This is because undefined4 is typedef'd as ypu realised. So we need to resolve those typedefs as much as possible. In any case we want to get rid of the `undefined...` variations so we do it like you just did it here?

gynt:
See fcc95ee9

Indeed pointer is not supposed to be exported as `pointer *`. I fixed that issue too.

src/OpenSHC/Globals/DAT_LoadingBarProgress.hpp
TheRedDaemon: grafik
Yes. Sadly, `uint` is also a typedef.

gynt: Resolved in f9f579d5a125d4357b6359bd8faf816f2358a6ba

gynt:
See f9f579d5a125d4357b6359bd8faf816f2358a6ba

gynt (approved):

PR #152, opened 2026-07-11
feature: add ai agent skills

TheRedDaemon:
Meant to help with support tasks to guide agents a bit better.
- Feeback on naming wanted
- "setup-function" could maybe just be a parsing script, although it would need to look into multiple files.
- "enable-local-function-dev" could also be a script, potentially.
- "find-function-implementation" might also be redundant in general, but the AI can receive guidance through our codebase this way.

Just testing around. There might be better way to do this. If we keep them, then can also be extended/changed. Especially the transforming scripts, should new unknown cases arise.

Creating an AGENTS.md is planned, as is a collection of MSVC structures I remember (I need to know where to put that one though, since it would be a more general list).
I would keep the scope of the PR to the skills for now, though.


.agents/skills/clean-decomp-body/SKILL.md
```diff
@@ -0,0 +1,153 @@
...
+
+### 2. Function Calls
+
+Replace direct calls:
+
+`Namespace::Function(args)` → `MACRO_CALL(Namespace_Func::Function)(args)`
+
+Replace member calls using global objects:
+
+`Namespace::Class::Function(&Global, args)` or `Namespace::Class::Function(Global, args)` → `MACRO_CALL_MEMBER(Namespace::Class_Func::Function, Global::ptr)(args)`
```
gynt: Or `Namespace::Class::Function(this, args)` ?

TheRedDaemon: Added additional point for member functions of the same class. Moved it under field transform. May not hurt to have the "determine class" part over the other, but no idea if this has any effect.

gynt (approved):

PR #155, opened 2026-07-13
refactor: Wilps contribution 2026-07-09

gynt:


src/OpenSHC/AI/AICState.hpp
```diff
@@ -120,31 +120,31 @@ namespace AI {
 
         uint aiShouldBuildPitchrig(PlayerID param_1);
 
-        BOOLEnum shouldNotBeBuildingHovel(PlayerID playerID, CommandBuildingType param_2);
+        BOOLEnum shouldNotBeBuildingHovel(PlayerID playerID, CommandBuildingTypeInt param_2);
```
TheRedDaemon, line 123: Before I try to look at 300 files: Is this intended, that the Enums are replaced here again?

TheRedDaemon (changes requested):

src/OpenSHC/AI/AICState.hpp
gynt: No, it isn't. I guess this was done on and old SARIF export before switched the enum stuff. Luckily I use python scripts to do that enum change so I can run it again and improve this PR.

TheRedDaemon: I do not know when you did the last sarif export, but after the renaming and this one, you should do another one.

PR #160, opened 2026-07-15
reimplement: SHC_3BB0A8C1_0x00449F10 100%

TheRedDaemon:
Now I only changed the parameter names, but maybe the naming could also be changed?

This function here is just another version of `playSFXAtLocation` without the volume part. It basically just pans the sound, but does not make it quieter or louder.


gynt:
Let's go with `playSFXAtLocationFullVolume`

gynt:
See f9f579d5a125d4357b6359bd8faf816f2358a6ba

gynt (approved):

PR #163, opened 2026-07-18
feat(scripts): add script to collect function data

TheRedDaemon:
Relies on a ton of regex. Since it is a support script, I am okey with it breaking in edge cases.
I just hope our general function declaration will not change anymore. 🙃 

EDIT: Added also a simple implemtation file setup script.
I also changed both skills. I am not 100 percent sure if I want this, since now they rely on the powershell scripts running.
But is saves tokens..
What do you think?

### DIFF
--- a/.agents/skills/openshc-find-function-implementation/SKILL.md
+++ b/.agents/skills/openshc-find-function-implementation/SKILL.md
@@ -14,7 +14,7 @@ Accepted formats:
 
 ## Resolution
 
-Powershell script call: `tools/reimplementation-helper/Get-Function-Data.ps1 -Partial -Query <query>`
+Powershell script call: `tools/reimplementation-helper/Get-Function-Data.ps1 -Partial -Query <input>`
 
 - If a function name was requested and the resulting name is different, mention the partial match.
 - Use the implementation file and the exists parameter.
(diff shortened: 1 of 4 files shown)


gynt:
I am wondering why adding a mcp server entry isn't easier. Then the skill can refer to this MCP entry point to do the stuff we want to save tokens.

TheRedDaemon:
I want to use the functions also manually, mainly. This is not only to aid the AI workflow.

I was thinking about making it at least python scripts, but:
- I do not like to call them via "python ..."
- I was premature optimizing by assuming that file scanning inside python should be slower than via a tool like "findstr". Do I have prove? Not at all. That being said, one would also able to call it via python.

Skill vs. MCP is a whole other debate. Some say a good CLI via skills is better for clear tasks, since it limits the initial token usage (since it discovers incremental) and might communicate with shorter returns.

Maybe this is pedantic.
Should I transform them to python scripts? They can then integrate with everything.

gynt:
Yes I think it is more robust in Python if it is to be used by humans and AI (also given that we support reimplementing from Linux). With that said, just start using the tooling and adapt it when needed! We can switch to Python/linux when that need is actually present

gynt (approved):

PR #173, opened 2026-07-24
reimplement: SHC_3BB0A8C1_0x0044ABB0 SHC_3BB0A8C1_0x0044A620 SHC_3BB0…

TheRedDaemon:
…A8C1_0x0044A6E0 100%

A good bunch of declarations need an update:
- `tweakVolume` has a parameter more.
- `DAT_SFX_Cooldown_WeAreUnderAttack` (0x00b986e4) requires a resolver
- `General_Warning10.wav` seems to be missing from the `string-literals`
- `DAT_00b986e0` requires a type change

Additionally, the float functions aside (that show up as mismatches, since we do not mark the internal functions currently):

Float and double constants seem to be interestingly handled. They seem to be not inline, but placed at addresses, and show up as mismatches due to this. No idea if it would be worth it to deal with this, though.


gynt:
See d444be0f and 69e56069

gynt (approved):

PR #174, opened 2026-07-26
reimplement: SHC_3BB0A8C1_0x0044AE90 100%

TheRedDaemon:
~Has a familiar issue:~
- ~One structure does not produce the fitting opcode structure.~
- ~Creating something that does messes up most other registers.~
- ~So either I require something to make the "messed up" situation work, or the initial mismatch might actually be related to some other pressure.~

Ha, not this time. Managed to pressure it into the right opcode. Feels a bit "wrong" code-wise, but whatever.


gynt:
See 1b97302f

gynt (approved):

PR #178, opened 2026-07-29
reimplement: SHC_3BB0A8C1_0x0044C740 100%

TheRedDaemon:
This will not compile while `DPERR.hpp` is added.

The rason is simple: It is an enum for MACROS from DirectPlay.
So either the enums need to be renamed, or (maybe better?) the enum should be removed and the direct play flags should be used.


gynt:
I stumbled into the same thing but forgot to update the export logic I think... I will triple check the DPERR situation.

gynt:
See 1b97302f

TheRedDaemon:
@gynt See changes: There were some leftovers of the type.

gynt:
See 81f69789c9746c47703c77083c28fa7e4e8c9a5d

gynt (approved):

PR #183, opened 2026-08-04
reimplement OpenSHC::IO::DecoderBuffer 100%

TheRedDaemon:


src/OpenSHC/IO/DecoderBuffer.hpp
```diff
@@ -15,9 +15,9 @@ namespace IO {
 
         using OpenSHC::IO::Helpers::BufferHelper;
 
-        uint __cdecl ReadBufferAndComputeHash(char* buf, uint* size, BufferHelper* param);
+        unsigned int __cdecl ReadBufferAndComputeHash(char* buf, unsigned int* size, void* param);
```
gynt, line 18: Is the change into an explicit `unsigned int` instead of `uint` intentional?

TheRedDaemon: I have the habbit to change things to the native types... of course, this can backfire should we ever want to go to 64bit... but this is something for the future...

Should I focus on using the types from `common.hpp`?

TheRedDaemon:
It is still missing some declaration changes.

src/OpenSHC/IO/DecoderBuffer.hpp
gynt: I understand the habit, but would be nicer if you use common.hpp because that is what ghidra only handles (uint)

gynt:
See 0611a372

gynt (approved):

PR #187, opened 2026-08-12
feat(compiler): set gl flags for specific files to remove mov ecx

### LINKED ISSUE #185: [OPTIMIZATION] Level of optimization applied in code across cpp files or across whole project?
Function at `OpenSHC/IO/DecoderState/encodeData.cpp` doesn't want to drop the redundant `mov ecx, offset` lines before repeated calls to copyData which doesn't clobber ECX (preventing a 100% match) unless:
In CMakeLists.txt:
set_source_files_properties(src/OpenSHC/IO/LowLevelMemory/copyData.cpp PROPERTIES COMPILE_FLAGS /GL)
set_source_files_properties(src/OpenSHC/IO/DecoderState/encodeData.cpp PROPERTIES COMPILE_FLAGS /GL)

or for the whole project /GL is set.

In case large .cpp files with multiple definitions were used note /Ob1 is required to not make the compiler inline methods (or declspec noinline)

Diving deeper, LowLevelMemory has two functions in which a lack of ECX clobbering has been used by the compiler in calls to these two functions. However, not all of the times the mov ecx is removed. Here are the cases (mov ecx removed) and the anti-cases (mov ecx not removed where it could/should have). Note that fillMemory_ByteValue has only one case in which mov ecx was removed. copyData has three cases.

If we assume that the /GL theory is applied to .cpp files of some classes and not others (perhaps they were even separate subprojects in a Visual Studio solution), then likely candidates for /GL are: LowLevelMemory, ResourceManager, DecoderState, WindowAndDirectDraw. Thematically, this makes sense as these things sound low-level and are closely tied to system/library IO.

Data:
```
copyData:
case: _HoldStrong::IO::ResourceManager::swapMapDataWithNextMap has subsequent mov ecx removed
case: _HoldStrong::IO::DecoderState::encodeData has subsequent mov ecx removed
case: _HoldStrong::IO::DecoderState::decodeData has subsequent mov ecx removed
anti-case: _HoldStrong::AI::AIVState::rotateAIV which should have had the second mov ecx removed
anti-case:  _HoldStrong::Map::WildlifeState::updateNofFpoints should have had the second mov ecx removed
anti-case: _HoldStrong::Map::Units::UnitsState::setDestinationForUnit should have had the second mov ecx removed

fillMemory_ByteValue:
anti-case: _HoldStrong::UI::Rendering::TextureRenderCore::clearSomeMemory 4/5 calls to fillMemory_ByteValue but mov ECX is repeated every time
anti-case: _HoldStrong::UI::Rendering::TextureRenderCore::loadGMFile
anti-case: _HoldStrong::Game::GameStateStructures::clearMapAndTimeAndPlayerData
anti-case: _HoldStrong::Game::GameStateStructures::clearSignpostData
anti-case: _HoldStrong::Game::GameStateStructures::clearDataAndSignpostDataIfNecessary
case: _HoldStrong::UI::Rendering::WindowAndDirectDraw::takeScreenshot mov ecx to subsequent call to copyData removed (cross function example!)
anti-case: _HoldStrong::Synchrony::GameSynchronyState::clearChatEvent
anti-case: _HoldStrong::Global::ClearPathFindingTileMaps
anti-case: _HoldStrong::Map::Navigation::PathFindingState::updateSeparateAreaTileMap
anti-case: _HoldStrong::AI::AICState::wipeAICMemory
anti-case: _HoldStrong::Map::LandscapeState::clearRocksAndTrees
anti-case: _HoldStrong::Map::TileMapState::resetMoatArray
anti-case: _HoldStrong::Map::TileMapState::resetPitchDitchArray
anti-case: _HoldStrong::Map::TileMapState::setupTileMapSections (in half of cases ECX is clobbered in between calls to fillMemory_ByteValue so the mov ecx is justified)
anti-case: _HoldStrong::Map::Units::TroopValueState::collectArcherUnitsByLocation
anti-case: _HoldStrong::Map::Units::TroopValueState::recomputeAttackInfo
anti-case: _HoldStrong::Map::Units::TroopValueState::aiRecomputeAttacks2
anti-case: _HoldStrong::Map::Units::TroopValueState::aiRecomputeAttacks
anti-case: _HoldStrong::Map::Units::UnitsState::setDestinationForUnit
```

gynt:
Resolves #185

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,14 @@ set_target_properties(OpenSHC.dll PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_B
 target_compile_definitions(OpenSHC.dll PRIVATE OPEN_SHC_DLL REIMPLEMENTED_CRT=0)
 target_precompile_headers(OpenSHC.dll PRIVATE $<$<COMPILE_LANGUAGE:CXX>:${PCH_FILE}>)
 
-# Set per file compiler flags
+# Set per file compiler flags, such as /GL
+# Some class methods do not clobber the ECX register and calling such methods repeatedly means
+# in assembly that a 'mov ecx, offset' instruction is ommitted. This appears to be difficult to reconstruct
+# in our reimplementation for two reasons: we have a one function per cpp file approach, which makes it more unlikely
+# that the compiler knows about the ECX optimization, and second, we don't know the original source code's file structure.
+# Perhaps they had multiple implementations per cpp file, but we need more cases to prove this is the case.
+# For now, we set the /GL flag for a select number of files. Setting the /GL flag in *both* the caller and callee function
+# implementation removes the unnecessary mov ecx instruction from the callers code.
 file_dependent_read_list("${CMAKE_SOURCE_DIR}/cmake/compiler-flags-gl.txt" COMPILER_FLAG_GL_FILES)
 set_source_files_properties(${COMPILER_FLAG_GL_FILES} PROPERTIES COMPILE_OPTIONS /GL)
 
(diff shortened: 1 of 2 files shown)


TheRedDaemon (changes requested):
Can you make it, through file naming and/or comment (likley comment), clear, why we are setting the GL flag for these files (LowLevelMemory + register optimization)?

Without the knowledge from our discussion, it would feel arbitrary otherwise.

TheRedDaemon (approved):

PR #196, opened 2026-08-22
reimplement OpenSHC::Map::Rocks 100%

gynt:
Easiest functions ever. Makes you wonder why they had rock types and update functions at all... Perhaps they had some kind of template for entities, units, buildings that also have these update functions.


TheRedDaemon (changes requested):
Two small questions, although, not much about the content:
- Do we have an GM enum or something like that, that fits these ids? If not, no problem, no need to create it now.
- It is ok here, but can you really remove the space in the signature from the exporter:
`void __cdecl Rocks ::UpdateRockFunc6()` 🙃?

TheRedDaemon (approved):

PR #199, opened 2026-08-22
Reimpl/simple functions set 1

gynt:


src/OpenSHC/Audio/mss/SoundSystem/markMusicChangePending.cpp
```diff
@@ -1,4 +1,4 @@
-#include "../SoundSystem.func.hpp"
```
gynt, line 1: My clang-format fixes this relative import to its absolute version. Why doesn't yours do the same @TheRedDaemon  and what do you prefer?

TheRedDaemon: We agreed on using the relative import to mark the "main" header file, so this is something that should stay.

No idea on your question, though. I still use the integrated clang version of Visual Studio Code. Did you update your version recently to the newest? Maybe something was added regarding the imports?

PR #210, opened 2026-08-28
reimplement: SHC_3BB0A8C1_0x004BC620 100%

TheRedDaemon:
That one should likely go to the other tutorial functions. Assuming they stay in UI or wherever they are?
Have not yet performed any change to the position.


gynt:
Moved it to ::UI::Helpers::

PR #211, opened 2026-08-28
reimplement OpenSHC::Input::MouseState 100%

TheRedDaemon:


src/OpenSHC/Input/Mouse/ClickTypeBitMaskEnumByte.hpp
```diff
@@ -13,7 +13,7 @@
 namespace OpenSHC {
 namespace Input {
     namespace Mouse {
-        typedef byte ClickTypeBitMaskEnumByte;
+        typedef char ClickTypeBitMaskEnumByte;
```
gynt, line 16: You sure? Changing this would mean changing all enum bytes to enum chars, which is fine I guess, since original enum is also usually int (not uint). Just want to be sure before I implement such a large change.

src/OpenSHC/Input/MouseState.hpp
```diff
@@ -46,7 +46,7 @@ namespace Input {
         undefined4 rightClickStart; // 0x0000003C length: 4
         BOOLEnum leftClickState; // 0x00000040 length: 4
         BOOLEnum midClickState; // 0x00000044 length: 4
-        BOOLEnum rightClickState; // 0x00000048 length: 4
+        int rightClickState; // 0x00000048 length: 4
```
gynt: You do use TRUE and FALSE in the .cpp files. Are you sure it should be an int?

TheRedDaemon: Since this is just an alias for and `int` it actually has no effect on the logic. I reverted the change, no idea why I did this.

src/OpenSHC/Input/Mouse/ClickTypeBitMaskEnumByte.hpp
TheRedDaemon: Yes. While a while, I clearly remember that I was hesitant and tested around, but it is a `char`.

PR #216, opened 2026-09-05
Reimplement native market pricing and goods trading

Krarilotus:
**TL;DR:** Reimplements the game's market prices and AI goods trading, preserving how purchases and sales change stored resources and gold. Adds a short wiki description of these rules.

The four C++03 implementations use the existing types and resolvers:

| Function | Original address | Raw reccmp score |
| --- | --- | --- |
| `GameStateStructures::getBuyPrice` | `0x004588A0` | 100% |
| `GameStateStructures::getSellPrice` | `0x00458910` | 100% |
| `AICState::buyGoods` | `0x004CC000` | 85.71% |
| `AICState::sellGoods` | `0x004CBFA0` | 81.82% |

Prices preserve division before multiplication. Purchases charge gold only after resource storage succeeds; sales preserve the order of gold/statistics updates, resource removal and trade display. Field access uses the struct resolver's `instance`; member-call arguments use `ptr`.

Rebuilt the full `OpenSHC.dll` with MSVC 2005 SP1 x86, CMake 3.31.6 and `RelWithDebInfo`, selecting these four sources locally. Compared each function against Crusader 1.41 SHA256 `3bb0a8c1e72331b3a30a5aa93ed94beca0081b476b04c1960e26d5b45387ac5a`. The trade diffs are confined to resolver calls and symbolic global references; control flow, register use and field accesses match. The soft status entries use 100.0% following the review guidance for these resolver-only differences; the table above retains the raw scores.

The comparison build uses a current Windows SDK manifest tool because the legacy `mt.exe` crashes on this machine. Compiler and optimization settings remain the project's; no build override is included in the PR. The DLL has not been deployed or tested in-game.

Overlap: `origin/reimpl/simple-functions-set-1` at `5b9da3d` already contains the two price implementations, with identical expressions. The AI trade routines are additional. These price files can be dropped when that branch lands; no competing implementation is intended.


src/OpenSHC/Game/GameStateStructures/getSellPrice.cpp
```diff
@@ -0,0 +1,14 @@
+#include "../GameStateStructures.func.hpp"
+
+namespace OpenSHC {
+namespace Game {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x00458910
+    int GameStateStructures::getSellPrice(int playerID, int resourceType, int amount)
+    {
+        // Preserve division before multiplication, including non-multiples of five.
```
TheRedDaemon: Comments like this can be removed. At least, I do not think this yields information beyond what is readable in the code.

src/OpenSHC/Game/GameStateStructures/getBuyPrice.cpp
```diff
@@ -0,0 +1,14 @@
+#include "../GameStateStructures.func.hpp"
+
+namespace OpenSHC {
+namespace Game {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x004588A0
+    int GameStateStructures::getBuyPrice(undefined4 playerID, int resourceType, int amount)
+    {
+        // playerID is unused; truncate the unit price before multiplying.
```
TheRedDaemon: Same as in `getSellPrice`. This only explains the logic under it.

src/OpenSHC/AI/AICState/sellGoods.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenSHC {
+namespace AI {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x004CBFA0
+    void AICState::sellGoods(int playerID, ResourceType resourceType, int amount)
+    {
+        int const reward = MACRO_CALL_MEMBER(Game::GameStateStructures_Func::getSellPrice, DAT_GameState::ptr)(
+            playerID, resourceType, amount);
+        DAT_GameSynchronyState::ptr->finalResults.finalGold[playerID] += reward;
```
TheRedDaemon: We support an `instance` access for the struct resolvers, which should be preferred in all cases that do not require the structs to be handled as pointers.

So instead of this:
```cpp
DAT_GameSynchronyState::ptr->finalResults.finalGold[playerID] += reward;
```
it would be this:
```cpp
DAT_GameSynchronyState::instance.finalResults.finalGold[playerID] += reward;
```
But function calls for example would still use:
```cpp
MACRO_CALL_MEMBER(... , DAT_GameState::ptr)(...)
```

This feature came later and I remember gynt updated the examples (I think, @gynt?), but some of the older reimplements still contain this.

docs/wiki.rst
```diff
@@ -29,6 +29,7 @@ For the recommended method of using SARIF files in Ghidra, see:
 The Game itself
 ------------------
 - :doc:`Load balancing of the core game engine <wiki/load-balancing-table>`
+- :doc:`Market pricing and goods trading <wiki/market-trading>`
```
TheRedDaemon: Did you replace something here?

docs/wiki/market-trading.md
```diff
@@ -0,0 +1,44 @@
+# Market pricing and goods trading
+
+These functions describe the native trading behavior used by automated market
+callers. They preserve the original game rules; callers remain responsible for
+deciding whether a trade should be attempted.
```
TheRedDaemon: Does `automated market` relates to the AI, or the auto market feature of the UCP?
The latter should not be mentioned here, since it should talk about the vanilla behaviour.

docs/wiki/market-trading.md
```diff
@@ -0,0 +1,44 @@
...
+`3bb0a8c1e72331b3a30a5aa93ed94beca0081b476b04c1960e26d5b45387ac5a`,
+using the project's MSVC 2005 SP1 x86 toolchain and RelWithDebInfo configuration.
+Both price functions match at 100% in reccmp. The trade functions score 81.82%
+(`sellGoods`) and 85.71% (`buyGoods`) with inactive resolvers. A separate
+instruction check confirmed that every non-call instruction is byte-identical
+and all six resolver call thunks target the corresponding original callees.
+
+This is a static binary comparison, not an in-game multiplayer test. Default
+source selection and resolver activation remain separate from these
+implementations.
```
TheRedDaemon: That is part of the process and not something that should be in the wiki.

docs/wiki/market-trading.md
```diff
@@ -0,0 +1,44 @@
...
+deciding whether a trade should be attempted.
+
+| Function | Crusader 1.41 address | Behavior |
+| --- | --- | --- |
+| `GameStateStructures::getBuyPrice` | `0x004588A0` | Returns `(buyPrice / 5) * amount`. |
+| `GameStateStructures::getSellPrice` | `0x00458910` | Returns `(salesPrice / 5) * amount`. |
+| `AICState::buyGoods` | `0x004CC000` | Attempts resource storage, then deducts gold and displays the trade on success. |
+| `AICState::sellGoods` | `0x004CBFA0` | Credits gold and final-result statistics, removes resources, then displays the trade. |
+
+## Details relevant to callers
```
TheRedDaemon: I think this part talks to much about the code. I am not sure it contains a lot more regarding the logic then the table above.
It also mentions the auto market again.

TheRedDaemon (changes requested):
Hi, thanks for the contribution!

Comments are on the code, also:

We have a "soft" status file [status/addresses-SHC-3BB0A8C1.txt](https://github.com/sourcehold/OpenSHC/blob/main/status/addresses-SHC-3BB0A8C1.txt) where we track the general status of the functions to avoid heavier test pipelines for the start.

The structure should be self explaining, just check some entries by searching for `Reimplemented`. The percentage also does not have to be too precise. If the only mismatches are the resolvers/proxies and the structure fits, saying "100.0%" is ok. If the structure still has issues, the percentage (with struct resolvers active) is ok.

Now I will also mention @gynt and add him to the review, since I will give him the final word on the wiki entry/purpose. Also, he has a lot of smaller functions in his branches and needs to check if they overlap with your MR.

src/OpenSHC/Game/GameStateStructures/getSellPrice.cpp
Krarilotus: Removed the redundant comment in 9edc084.

src/OpenSHC/Game/GameStateStructures/getBuyPrice.cpp
Krarilotus: Removed in 9edc084; the expression is unchanged.

src/OpenSHC/AI/AICState/sellGoods.cpp
Krarilotus: Updated both trade functions in 9edc084 to use `instance` for field access and retained `ptr` for member calls. Rebuilt and compared the DLL; the remaining differences are resolver references/calls.

docs/wiki.rst
Krarilotus: No existing entry was replaced: the diff adds only the market-trading link below the load-balancing page. The rest of the index is unchanged.

docs/wiki/market-trading.md
Krarilotus: Clarified that this describes the vanilla AI in 9edc084 and removed the UCP Automarket references from the wiki.

docs/wiki/market-trading.md
Krarilotus: Removed the verification section from the wiki in 9edc084. Build and comparison evidence is now in the PR description.

docs/wiki/market-trading.md
Krarilotus: Shortened this to the price rounding, storage-failure behavior and gold/statistics effects in 9edc084. Removed the decompiler/register discussion and Automarket references.

Krarilotus:
Updated in 9edc084: field access now uses `instance`, the redundant comments are removed, and the wiki describes vanilla trading behavior without build-process details. I also added the four soft status entries. Fresh RelWithDebInfo DLL comparisons give 100% for both price functions; the trade routines retain the resolver-only differences documented in the PR, so their status follows the suggested 100.0% convention.

I checked Gynt's `reimpl/simple-functions-set-1` branch as well: it already has both price expressions. The AI buy/sell routines are additional; I have noted the overlap in the description so the price files can be dropped when that branch lands.

docs/wiki.rst
gynt: I think it is unclear that this article will be about AI. Please add a AI subsection and add this entry to it with AI in the title. And make sure the .md file refers to AI market behavior explicitly.

gynt (changes requested):

TheRedDaemon (dismissed):
Looks good from my side.

docs/wiki.rst
Krarilotus: Updated in 5941a84: the wiki now has an AI Behavior subsection, and both the link and article title explicitly say AI market pricing and goods trading. The introduction and trade descriptions now identify the AI routines and their callers.

gynt (approved):

PR #217, opened 2026-09-05
Reimplement stockpile footprint cleanup (native byte match)

Krarilotus:
**TL;DR:** Recreates the original cleanup of a stockpile's nine walkable tiles, including the building's rubble behavior. The implementation uses existing map and building structures and matches the original function at 100% in reccmp.

Reimplements `TileMapState::clearStockpileFootprintTiles` at `0x004FAF70`. Each tile loses the existing `L_STOCKPILEUnk` and `L_WALL_OR_GATEHOUSE` logic flags, regains its default height, follows its referenced building's `noRubble` branch, and finally clears `AlphaGFXLayer`.

`MiscDisplayLayer` is a separate ushort layer without a corresponding flag enum in the current headers. Its `0x4000` value remains literal until its meaning is identified. The wiki documents this limitation and the ownership of the map layers and global inputs.

Validation: full x86 RelWithDebInfo DLL build with VS2005 SP1 passes after the flag-name change; linked reccmp still reports **100%** at `0x004FAF70`. The local build uses a current Windows SDK manifest tool because the legacy tool crashes. No DLL deployment or gameplay test was performed. The emulation script was removed from this PR as requested in review.


tools/verification/verify_stockpile_cleanup.py
```diff
@@ -0,0 +1,81 @@
+"""Compare a VS2005 /O2 DLL-mode object with original SHC 1.41 under x86 emulation.
```
TheRedDaemon: As mentioned in other places: Please remove the emulation scripts if you matched 100%. They are optional in case you did not.

src/OpenSHC/Map/TileMapState/clearStockpileFootprintTiles.cpp
```diff
@@ -0,0 +1,31 @@
...
+    // FUNCTION: STRONGHOLDCRUSADER 0x004FAF70
+    void TileMapState::clearStockpileFootprintTiles(int x, int y)
+    {
+        for (int i = 0; i < 9; ++i) {
+            int tile = DAT_ViewportRenderState::instance
+                           .translationMatrix[y + DAT_TerrainDefinedData::instance.StockpilePathableOffsets[i].y]
+                           .addXgetTile
+                + x + DAT_TerrainDefinedData::instance.StockpilePathableOffsets[i].x;
+
+            this->LogicLayer[tile] &= ~0x102;
```
TheRedDaemon: Are these two flags? Are there documented values for these logic values? @gynt?

src/OpenSHC/Map/TileMapState/clearStockpileFootprintTiles.cpp
```diff
@@ -0,0 +1,31 @@
...
+            this->HeightLayer[tile] = this->DefaultHeightLayer[tile];
+            if (DAT_BuildingsState::instance.buildings[this->AlphaGFXLayer[tile]].noRubble == 0) {
+                this->BuildingWasLayer[tile] = 0;
+            } else {
+                this->MiscDisplayLayer[tile] |= 0x4000;
```
TheRedDaemon, line 25: Same here: It would be interesting if we already have enums for this, at least if they are all flags.

TheRedDaemon (changes requested):

tools/verification/verify_stockpile_cleanup.py
Krarilotus: Removed in ba14b9b. The full DLL rebuild and linked reccmp comparison still give 100% at 0x004FAF70; the PR now contains the implementation and its behavior documentation.

src/OpenSHC/Map/TileMapState/clearStockpileFootprintTiles.cpp
Krarilotus: Yes: 0x102 combines the existing Logic1 flags L_STOCKPILEUnk (0x2) and L_WALL_OR_GATEHOUSE (0x100). ba14b9b uses those names from LogicHelpers/Logic1.hpp. The rebuilt function retains its 100% match.

src/OpenSHC/Map/TileMapState/clearStockpileFootprintTiles.cpp
Krarilotus: I checked the current headers: MiscDisplayLayer is a ushort array and I could not find a flag enum for that layer. I kept 0x4000 literal and documented the unresolved meaning. A same-valued Logic1 flag belongs to another layer, so reusing that name here would be misleading; identifying this bit needs its own consumer trace.

docs/wiki/stockpile-footprint-cleanup.md
```diff
@@ -0,0 +1,35 @@
...
+
+The map layers belong to the supplied `TileMapState` instance. Building
+metadata, terrain offsets and row translation come from the corresponding
+global game structures.
+
+These logic names come from the existing `Logic1` enum; their combination here
+does not establish additional stockpile semantics. `MiscDisplayLayer` is a
+separate `ushort` layer with no corresponding flag enum in the current headers.
+Its `0x4000` bit is kept literal pending identification of its consumers; a
+same-valued flag from another layer would not establish its meaning.
```
TheRedDaemon: This is far to technical again. That extends to the wiki entry itself.
I personally think the other text part are also too technical in the way that they talk a bit much about what the code does.
Do not get me wrong, these entries can state stuff like the three points, like "it clears that and that flag on the map logic layer and restores the height", but things like the part after, talking about where these infos come from, are pure code information and provide no information about what the game does on a game logic level.

src/OpenSHC/Map/TileMapState/clearStockpileFootprintTiles.cpp
TheRedDaemon: Thats fine. I assume some flags will end up being not fully understood even long after we are done with the reimplementation, and this will already take quite some time.

TheRedDaemon (changes requested):

docs/wiki/stockpile-footprint-cleanup.md
Krarilotus: Rewrote the article in 4596df5 around the visible game behavior: clearing the walkable tiles, updating rubble/display information, and the shared placement rules. Removed addresses, enum provenance and discussion of how the evidence was obtained. The unidentified display effect remains stated briefly without guessing its meaning.

TheRedDaemon: Looks a lot better.
Two small things I still noticed:
- The file is called "stockpile-footprint-cleanup.md", yet half is about the placement. Would this better be a file about the stockpile?
- This one is not linked like some in the other PRs, or not? Also, while it does not need to happen now, putting these all into the wiki root could might become a hassle later if there are a lot of wiki files.

Krarilotus:
Renamed the article to `game-mechanics/stockpile.md` so its title covers both placement and cleanup, and linked it from the wiki overview. The game-mechanics directory also gives these gameplay articles a place outside the wiki root. Sphinx generated the renamed page and link; no implementation code changed.

docs/wiki/stockpile-footprint-cleanup.md
TheRedDaemon: Looks good

TheRedDaemon:
@Krarilotus LGTM

Please resolve the conflict. We usually do this via rebasing, basically, if you are on your branch, you would rebase on "main" and force push the result onto your branch.

PR #220, opened 2026-09-05
Reimplement multiplayer command ownership and document network flow

Krarilotus:
**TL;DR:** Reimplements how the game identifies the player behind a multiplayer command. The accompanying article explains how command delivery and synchronization fit together.

`GameSynchronyState::translateMultiplayerIDsIntoPlayerIDs` returns the local slot in either single-player mode. In other modes it checks slots 1–8, retaining the last match and returning zero if none matches. It does not change game state.

The implementation now follows TheRedDaemon's review: two separate early returns and a simple loop. A fresh MSVC2005 SP1 `/O2` DLL build and linked reccmp comparison report **100% at Crusader 0x0047EAF0**. The status entry is updated; the optional emulation scripts have been removed from the PR now that the function matches. No live DLL deployment was performed.

The wiki describes the surrounding timed-command, immediate-command and synchronization responsibilities. Those larger routines are documented analysis, not additional reimplementations in this PR.


Krarilotus:
﻿Added native system-message evidence to the linked network-flow note in 8a9c79d. The player-removal branch calls the identity translator at Crusader 0x490755, then immediately calls removePlayerFromLobby at 0x49075B without reloading ECX: another concrete caller for the receiver-preservation requirement in this implementation.

The note also distinguishes host timing/hash resets from timed-command dispatch and records the system type constants against the SDK header. This update changes documentation only; the C++ implementation and its previously reported 6,156 differential comparisons are unchanged. The larger receive routine is analyzed, not claimed newly reimplemented.

src/OpenSHC/Synchrony/GameSynchronyState/translateMultiplayerIDsIntoPlayerIDs.cpp
```diff
@@ -0,0 +1,36 @@
...
+            player = 4;
+        if (this->currentPlayerFullIDArray[5] == playerHandle)
+            player = 5;
+        if (this->currentPlayerFullIDArray[6] == playerHandle)
+            player = 6;
+        if (this->currentPlayerFullIDArray[7] == playerHandle)
+            player = 7;
+        if (this->currentPlayerFullIDArray[8] == playerHandle)
+            player = 8;
+        return player;
```
TheRedDaemon, line 25: I checked this myself, since functions of this size can usually be brute forced and are good to learn stuff.

Here is the body I managed to match:

```cpp
uint player = 0;
if (this->currentGameMode == Game::GM_SOLITARY) {
    return this->currentPlayerSlotID;
}

if (this->currentGameMode == Game::GM_SKIRMISH_SINGLE_PLAYER) {
    return this->currentPlayerSlotID;
}

for (int i = 1; i < 9; ++i) {
    if (this->currentPlayerFullIDArray[i] == playerHandle) {
        player = i;
    }
}
return player;
```

The loop does actually not contribute to the matching value. Cases that show these strongly repeating patterns with clear "run" values, like increasing ints, are sometimes unrolled loops.

The not matching part was actually the early return. Here the knowledge was in the assembly. Such chained logic tests like in your version usually resolve to two checks after another. In this case however, the first check in the originals assembly was directly followed by the early return body, then comes the second check that actually jmps back to the firsts return body, a sign that these two where different early returns.

I recommend our cheat sheet in the project root. It is not the best read stylewise and I should clean it a bit one day, but it contains a lot of weird patterns we noticed in the compiler behaviour.

TheRedDaemon (changes requested):

src/OpenSHC/Synchrony/GameSynchronyState/translateMultiplayerIDsIntoPlayerIDs.cpp
Krarilotus: Applied the separate early returns and the loop you suggested. A fresh MSVC2005 SP1 DLL build now gives 100% in linked reccmp at 0x0047EAF0. I also removed the optional emulation scripts now that the function matches. Thanks for explaining the return-block ordering; the manual unrolling was unnecessary.

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+   mode. Timed payloads go into the ring. Commands with signed time <= 0 copy to
+   the fixed parameter area and execute immediately at `0x480425`, then clear
+   their temporary ring entry. They never reach `processWaitingCommands`.
+5. `getCommandIDFromCommandSelectionStuff` (`0x480440`) selects due, unprocessed
+   entries, with a batch cap of 100. It translates sender handles and stably
+   orders the selected entries by logical player slot. A recorder must observe
+   actual dispatch order; transport arrival order is insufficient.
+6. `processWaitingCommands` (`0x4892F0`) translates each sender again, sets the
+   execution action/parameter state, calls the handler and marks the entry
+   processed. The translator is `0x47EAF0`, reimplemented by this contribution.
```
TheRedDaemon, line 34: ` The translator is `0x47EAF0`, reimplemented by this contribution.`
That needs to be removed. The wiki has nothing to do with the function you currently implement.

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+
+The timestamp is thus 24-bit on the wire and 32-bit in the ring. The examined
+receive routine does not reconstruct a higher epoch. Long-duration recording
+and replay must preserve that distinction rather than assuming identical wire
+and internal formats. This observation does not establish a reproduced wrap bug.
```
TheRedDaemon, line 39: `Long-duration recording and replay must preserve that distinction`

Did something UCP module related slip through here?

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+
+After a successful `IDirectPlay4A::Receive`, the receiver compares the sender
+with zero (`DPID_SYSMSG`). Its system-type switch begins at `0x490735`
+(Extreme `0x490895`). Failed receives and ordinary nonzero senders branch away
+before this point. Values below also agree with the DirectX SDK `dplay.h`:
+
+- `DPSYS_DESTROYPLAYERORGROUP` (`5`) reads the handle at message offset 8 and
+  calls the identity translator at `0x490755`. The next call is
+  `removePlayerFromLobby` at `0x49075B`, without reloading ECX. This is another
+  concrete native caller requiring this reimplementation to retain ECX.
```
TheRedDaemon, line 51: The `ECX` part is interesting for the reimplementation, but not for the wiki.

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+- `DPSYS_CREATEPLAYERORGROUP` (`3`) and `DPSYS_SESSIONLOST` (`0x31`) fall through
+  to the next receive iteration without special handling in this switch. This
+  describes the original routine; it does not imply that a transport or replay
+  implementation may disregard the broader connection lifecycle.
+
+Polling roster and synchronization fields at simulation boundaries cannot
+establish that no system event occurred between them. Recorder diagnostics can
+observe the type-switch entry before mutation, but replaying those events still
+requires their semantics and timing. DirectPlay system structures can contain
+process pointers; copying their bytes is not a portable replay format.
```
TheRedDaemon, line 66: Replay talk.

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+`sendPendingResyncCommandsInBudget` (`0x48E680`) sends mismatching sections in
+bounded batches, retaining its category/item position between calls. These
+include units, buildings, player data and tile-map regions. Resync replaces
+simulation data; treating every immediate command as presentation-only is wrong.
+
+A useful recorder must account separately for timed command execution, immediate
+state-changing commands, roster/host transitions and resync transfers. Reliable
+transport delivery alone cannot restore missing replay state or make
+frame/audio-driven RNG calls deterministic. An extension's ordinary save
+integration also does not demonstrate integration with native resync transfers.
```
TheRedDaemon, line 89: This whole part:
There is some general information about the technical side in here, but most is about replay.

docs/wiki/multiplayer-command-flow.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Validation scope
+
+The accompanying player-identity checker compiles and executes the actual C++
+function against the originals in both variants. It covers last-match semantics,
+sentinels, missing handles, relocated receivers and the retained ECX value.
+Other routines in this note were inspected in the named Ghidra project and
+original assembly; they are not newly reimplemented or live-validated here.
+See [reproduction instructions](../../tools/reimplementation-tests/PLAYER-IDENTITY.md).
```
TheRedDaemon, line 98: That is likely a leftover?

TheRedDaemon (changes requested):
The wiki entry is very technical, but since it relates to the very complex multiplayer behaviour, I think it is ok.
One day this could link to the object definitions to make it more understandable, but currently one might just end up creating old references should something still move.

---

Also, should you not do it, you really need to read everything yourself. You can always split notes for your modules out, but the assessments regarding them are not interesting here in this wiki.
It also helps on another level, if it starts to annoy you while reading and you think it is too much, then it might actually be 🙂 .

PR #222, opened 2026-09-06
Reimplement the immediate synchronization hash command

Krarilotus:
**TL;DR:** Reconstruct the immediate command that exchanges player synchronization information.

This native function sends and receives lag, hash and match time for the existing synchronization checks.

**Changes:** Implement `CommandCheckSync` at `0x480B10` using the existing serializers/resolvers and the original ten-byte payload.

**Review / testing:** DLL build and native comparison cases passed; linked reccmp is 75.47%. The 203-byte native-bound comparison is a separate measurement. **TheRedDaemon requested a status-file update, and the current PR still changes only the C++ file.** Complete that review request before merging.

### DIFF
--- a/src/OpenSHC/Synchrony/Commands/CommandCheckSync.cpp
+++ b/src/OpenSHC/Synchrony/Commands/CommandCheckSync.cpp
@@ -1,0 +1,64 @@
+#include "../Commands.func.hpp"
+
+#include "OpenSHC/Synchrony/GameSynchronyState.func.hpp"
+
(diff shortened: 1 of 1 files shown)


Krarilotus:
The build limitation is resolved: the normal full `OpenSHC.dll` target now passes using the existing MSVC2005 SP1 compiler and a Microsoft-signed Windows SDK manifest tool. All 2,560 original serializer/command cases still pass; the C++ commit is unchanged and no DLL was deployed. The description now reflects the successful full build.

TheRedDaemon (changes requested):
It is missing the status file update.

PR #223, opened 2026-09-06
Reimplement host synchronization checks and periodic messages

Krarilotus:
**TL;DR:** Reconstruct host synchronization checks and periodic sync messages.

OpenSHC needs readable C++ counterparts for the native host comparison, periodic sender and unit-buffer hash helper.

**Changes:** Implement the three functions through existing resolvers, retaining original timing, peer filtering, receiver identity and acknowledgements.

**Review / testing:** DLL build and 9,380 differential cases were reported passing. Linked reccmp scores are 98.02%, 92.59% and 60.00%. **Outstanding maintainer requests:** update status, use explicit `this->`, remove matching validation files and the redundant validation README. The current diff still contains the validation directory and no status update. The native-bound byte match does not replace those requests.

### DIFF
--- a/src/OpenSHC/Synchrony/GameSynchronyState/checkGameSync.cpp
+++ b/src/OpenSHC/Synchrony/GameSynchronyState/checkGameSync.cpp
@@ -1,0 +1,59 @@
+#include "../GameSynchronyState.func.hpp"
+
+#include "OpenSHC/Globals/DAT_GameSynchronyState.hpp"
+
+namespace OpenSHC {
+namespace Synchrony {
+    // FUNCTION: STRONGHOLDCRUSADER 0x0048CB00
+    void GameSynchronyState::checkGameSync()
+    {
+        int referenceHash = 0;
+        int referenceTime = 0;
+        if (DAT_HashCountdown == 0) {
+            DAT_GameHalted = 0;
+            for (int player = 1; player < 9; ++player) {
+                if (currentPlayerFullIDArray[player] != -1 && unknownPlayerInfo_03[player] == 0) {
+                    referenceHash = HASH_HashTotal[player];
+                    referenceTime = DAT_PlayerMatchTimes[player];
+                    break;
+                }
+            }
+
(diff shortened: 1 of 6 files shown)


src/OpenSHC/Synchrony/GameSynchronyState/checkGameSync.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "OpenSHC/Globals/DAT_GameSynchronyState.hpp"
+
+namespace OpenSHC {
+namespace Synchrony {
+    // FUNCTION: STRONGHOLDCRUSADER 0x0048CB00
+    void GameSynchronyState::checkGameSync()
+    {
+        int referenceHash = 0;
+        int referenceTime = 0;
+        if (DAT_HashCountdown == 0) {
```
TheRedDaemon, line 12: You might be able to increase the readablility by switching this to an early return (if it does not break the byte matching).

src/OpenSHC/Synchrony/GameSynchronyState/checkGameSync.cpp
```diff
@@ -0,0 +1,59 @@
...
+            DAT_GameHalted = 0;
+            for (int player = 1; player < 9; ++player) {
+                if (currentPlayerFullIDArray[player] != -1 && unknownPlayerInfo_03[player] == 0) {
+                    referenceHash = HASH_HashTotal[player];
+                    referenceTime = DAT_PlayerMatchTimes[player];
+                    break;
+                }
+            }
+
+            if (DAT_GameSynchronyState::instance.isHost && flag_0xbec == 0) {
```
TheRedDaemon, line 22: Same here.

TheRedDaemon (changes requested):
- Please add the status file updates.
- Also, we prefer explicit `this->` accesses to member variables.
- Should the files match, please remove the validation files.
- The readme is also not fitting, the partial explanation of the emulation validation would be naturally part of the agent skill. The rest is directly bound to this function and redundant.

PR #224, opened 2026-09-06
reimplement OpenSHC::Text::TextEditorState

TheRedDaemon:
I noticed it while reimplementing `drawBorderStyle0x20`:
`dialogContentWidth` and `dialogContentHeight` are likely swapped.
And if not them, then `dialogWidth` and `dialogHeight`.
It should be fixed within this PR, I am just unsure whether or not it should be done now or at the end.

`intArray1` is seems to be an array of `wchar_t*`.

If I forget it, please remember me.

---

Many of the shorter functions share knowledge with some bigger and certain Menu functions.
These might partially all have been one giant file. Once the day comes, we likely need to use the GL method again. It is important to be careful then, so that we do not end up giving files knowledge of each other that they should not have.

---

There seem to be `ints/shorts/wchar_ts` that represent certain help text tokens. These could use an enum.

---

The L-strings need to be replaced with a string file reference.

---

Document the hlp-file structure a bit.


src/OpenSHC/OS.func.hpp
```diff
@@ -171,5 +171,13 @@ namespace OS_Func {
         void(__cdecl*)(int _Code), REIMPLEMENTED_CRT, Address::SHC_3BB0A8C1_0x00583D55, &OpenSHC::OS::_exit)
     _exit;
 
+    MACRO_FUNCTION_RESOLVER(int(__cdecl*)(FILE*, wchar_t const*, ...), REIMPLEMENTED_CRT,
+        Address::SHC_3BB0A8C1_0x005807A8, &OpenSHC::OS::_fwprintf)
+    _fwprintf;
+
+    MACRO_FUNCTION_RESOLVER(
+        long(__cdecl*)(wchar_t const*), REIMPLEMENTED_CRT, Address::SHC_3BB0A8C1_0x00580BFA, &OpenSHC::OS::__wtol)
```
gynt: 0x00580c0b is the thunk containing the jmp, so that one is actually the public-facing one, right? Or do you thing it should be 0x00580BFA

PR #226, opened 2026-09-08
Reimplement tick orchestration and document game-loop ownership

Krarilotus:
**TL;DR:** Reconstruct the tick coordinator while preserving native pause and update order.

A call to the tick coordinator is not always a completed simulation step; paused maintenance and synchronization can take different paths.

**Changes:** Implement `GameStateStructures::processGameTick` at `0x45CD10` with existing resolvers. Add a short game-loop reference and the soft status entry.

**Review / testing:** DLL build passed. Structural comparison resolves the 54 helper targets and finds one unused ECX setup difference; linked raw reccmp remains 35.98% and soft status is 99.0%. Those are different measures, not a claim of identical DLL bytes. No live deployment was reported.


src/OpenSHC/Game/GameStateStructures/processGameTick.cpp
```diff
@@ -0,0 +1,184 @@
...
+            if (!MACRO_CALL_MEMBER(GameCore_Func::isGameHaltingMenuOpen, DAT_GameCore::ptr)()) {
+                MACRO_CALL_MEMBER(Random::RNG_Func::nextRandomNumber2, SEC_RNG::ptr)();
+                MACRO_CALL_MEMBER(Random::RNG_Func::nextRandomNumber1, SEC_RNG::ptr)();
+                ++DAT_GameCore::instance.mapTimeInTicks;
+                MACRO_CALL_MEMBER(Game::GameStateStructures_Func::processSingleTimeTick, DAT_GameState::ptr)();
+            }
+        }
+        if (static_cast<int>(DAT_TileMapState::instance.DAT_FutureMapOrientation) < 8) {
+            DAT_RotateMapOrPullDownTerrain::instance = 1;
+        }
```
TheRedDaemon, line 109: Is the remaining mismatch related to this goto?

If yes and you not have tried it, you could invert the if block and pack the following block inside.
Naturally, this would mean if `static_cast<int>(DAT_TileMapState::instance.DAT_FutureMapOrientation) < 8` would be tested (since the if is inverted) and this would be false, then compiler would know that it can skip the other if-block before the jmp target and you can avoid the goto.

TheRedDaemon (changes requested):

PR #227, opened 2026-09-10
Reimplement adjacent tile region and height check

Krarilotus:
**TL;DR:** Reconstruct the native neighboring-tile region and height check.

Moat selection uses this helper to ask whether a neighboring tile belongs to an area within the native height limit.

**Changes:** Implement `TileMapState::findTileInSameAreaAndNoTooHeightDifference` at `0x500370`: examine eight offsets, allow at most 16 units upward and compare the signed region ID.

**Review / testing:** DLL build and 100% effective reccmp were reported; two independent setup instructions have different ordering. **TheRedDaemon requested the matching status-file update, which is still absent from the current diff.** This reconstructs existing behavior rather than changing moat routing.

### DIFF
--- a/src/OpenSHC/Map/TileMapState/findTileInSameAreaAndNoTooHeightDifference.cpp
+++ b/src/OpenSHC/Map/TileMapState/findTileInSameAreaAndNoTooHeightDifference.cpp
@@ -1,0 +1,24 @@
+#include "../TileMapState.func.hpp"
+
+namespace OpenSHC {
+namespace Map {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x00500370
+    BOOLEnum TileMapState::findTileInSameAreaAndNoTooHeightDifference(int area, int tile, int row)
+    {
+        int direction = 0;
+        int* neighborOffset = this->directionTranslationMatrix[row];
+        for (; direction < 8; ++direction, ++neighborOffset) {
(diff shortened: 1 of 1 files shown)


src/OpenSHC/Map/TileMapState/findTileInSameAreaAndNoTooHeightDifference.cpp
```diff
@@ -0,0 +1,24 @@
...
+namespace OpenSHC {
+namespace Map {
+
+    // FUNCTION: STRONGHOLDCRUSADER 0x00500370
+    BOOLEnum TileMapState::findTileInSameAreaAndNoTooHeightDifference(int area, int tile, int row)
+    {
+        int direction = 0;
+        int* neighborOffset = this->directionTranslationMatrix[row];
+        for (; direction < 8; ++direction, ++neighborOffset) {
+            int neighborTile = tile + *neighborOffset;
```
TheRedDaemon, line 12: Can you check if this loop would also work by index access and iterating over the directions?
Basically a normal for-loop and then using `this->directionTranslationMatrix[row][direction]`?
If not might `neighborOffset[direction]` possible, avoiding the pointer?

TheRedDaemon (changes requested):
This is missing the matching status file update.

